PR #42, opened 2014-03-13
polish translations: minutes

samupl:
In polish translations, there seem to be missing a '%' character. When I lock my screen and try to unlock it, the time is (for example), 09:M (obviously, M should be replaced by current minutes).


NiceandGently:
@stefano-k 
please merge to 1.8,
see https://bugzilla.redhat.com/show_bug.cgi?id=1199754

Issue #54, opened 2014-07-29
Desktop temporary visible when returning from suspend-to-disk 


SanTortilla:
System: Up to date Mint Mate 17 x64 with Mate 1.8.1
Priority: Medium to High (data security)

Issue: When using suspend-to-disk (S4) serval times, sometimes when the system restarts the desktop with all open programs and data is visible for 2-3 seconds. After 2-3 sec. the lockin-window appears.

What should happend: The lockin-window should appear initially.

Whats the problem: Everybody how starts my pc/notebook is able to have a (short) look at my private data.

If mate-screensaver isn't responsible for locking the screen, please move this issue to the correct mate-module.

NiceandGently:
@SanTortilla 
Which DM are you using?
And do you have another DM installed?

SanTortilla:
My DM is MDM (installed by default). I havent installed any other DM.

NiceandGently:
Ok, may hope died that your issue is related to https://github.com/mate-desktop/mate-screensaver/commit/269333e2683e6dd9975f298267cac177b889541a

paravoid:
This is a long-standing bug of gnome-screensaver that has been fixed by distribution patches in at least Debian/Ubuntu (e.g. see https://bugs.launchpad.net/ubuntu/+source/xorg-server/+bug/546578/comments/97).

The reason for this is when the screensaver initiates a screen fade before locking. When suspending, there is a race between actually suspending the system and the fade finishing.

I'd consider this an important security issue. The fix that Ubuntu/Debian have applied is trivial: in `src/gs-manager.c` there is this code:
```
        /* fade to black and show windows */
        do_fade = TRUE;
```

Setting this to FALSE is enough for fixing this.

infirit:
If we set this then there is no point in keeping the fade code around.  I am not sure we are ok with disabling fading like that.

I did push a [commit](https://github.com/mate-desktop/mate-screensaver/commit/599bc816e0dca82f921e6d77c2018d7818e7b7b2) that considerably reduces the time it takes when explicitly asked to lock. It is fast enough that I can not really see the fade happening when running ``mate-screensaver-command --lock``. This should be fast enough for most cases but if ditro's want to disable it completely than they can patch it like they did with gnome.

The fade due to inactivity is not affected by that commit.

paravoid:
I'll try it. If it's fast enough that you don't see the fade, though, what's the point of having it?

[...]

PR #79, opened 2015-09-14
Don't fade to black when locking the screen

paravoid:
Fading to black has negative security consequences: the screen locking
and actual suspend race with each other, and this can result in the
screen's contents being briefly visible after resume.

This has been reported as #54 but has been known and patched for in
Debian/Ubuntu since at least 2010 (see LP 546578).


clefebvre:
https://github.com/linuxmint/cinnamon-screensaver/blob/master/src/gs-manager.c#L1212
We've been running with that for a while in cinnamon-screensaver. Definitely worth losing a bit of polish to ensure things don't go wrong, this is quite critical.

PR #80, opened 2015-09-14
Fix broken systemd support

paravoid:
The current state of support for systemd is completely broken — this fixes two things:
1) Actually handle Lock/Unlock etc. events for systemd; despite most of the code for it being there, it's never actually activated(!)
2) Handle the PrepareForSleep systemd event so that the screen actually locks when the system suspends by e.g. systemctl suspend.

The first one is probably the fix to mate-desktop/mate-session-manager#27 as well; I don't think this was actually a session manager bug.

Both of these probably belong in the 1.10.x branch as well.

### DIFF
--- a/src/gs-listener-dbus.c
+++ b/src/gs-listener-dbus.c
@@ -1541,21 +1541,10 @@ _listener_message_path_is_our_session (GSListener  *listener,
 		return FALSE;
 
 	if (listener->priv->session_id == NULL)
-	return FALSE;
+		return FALSE;
 
-#ifdef WITH_SYSTEMD
-	/* The bus object path is simply the actual session ID
-	* prefixed to make it a bus path */
-	if (listener->priv->have_systemd)
-		return g_str_has_prefix (ssid, SYSTEMD_LOGIND_SESSION_PATH "/")
-			&& strcmp (ssid + sizeof (SYSTEMD_LOGIND_SESSION_PATH),
-			listener->priv->session_id) == 0;
-#endif
-
-#ifdef WITH_CONSOLE_KIT
 	if (strcmp (ssid, listener->priv->session_id) == 0)
 		return TRUE;
-#endif
 
 	return FALSE;
 }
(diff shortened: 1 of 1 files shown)


infirit:
I don't run systemd any-more so I am hoping my logind friends can look at this and have a go :smile:.

monsta:
Would one be able to test it on a system which has ```sysvinit``` but ```/lib/systemd/systemd-logind``` is running during the session?

raveit65:
Thank you, this works well in fedora 22
What i tested:
loginctl lock-session 1 -->  lockscreen pulls in
loginctl unlock-session 1 --> lockscreen pulls in
suspend on lid-close --> lockscreen pulls in
suspend after a timeout --> lockscreen pulls in

Hibernate is untestable because there is a general issue with systemd here, the box starts always a new system boot.
But it seems that it improve also the missing locksreen with user-switching in fedora.
In my test with 2 user accounts 1 account is always locked the other only after the screensaver pulls in. W/o the patch a happy user switching via crtl-alt /F1/F2 was possible w/o a lockscreen until screensavers pulls in.
But that's another story and maybe a fedora specific issue.
So, i see no regressions and only a big improvement.

....thank you again

@flexiondotorg 
Please can you test and merge it to master and 1.10.
I need that really for forthcomming f23 release.

raveit65:
> Would one be able to test it on a system which has sysvinit but /lib/systemd/systemd-logind is running during the session?
I can't imagine that a distro do this mixing horror of complete different systems ;)

monsta:
Hmm, ```loginctl unlock-session <number>``` from system console (Ctrl-Alt-F1) while lockscreen is active causes the session to be unlocked right away, bypassing password prompt. Is that ok?

raveit65:
Only if you're root on second console ;)
Normal user can't unlock his own or other locked sessions.

monsta:
Yes, I logged in as root there.

monsta:
@clefebvre: the first commit seems to be based on https://github.com/linuxmint/cinnamon-screensaver/pull/36. I see you've posted some remarks in the comments there. Do we need something else here?

clefebvre:
@monsta, I'm not sure. Note that the cinnamon-screensaver PR was merged all the same. I'm not ready to test power management just yet, but if raveit and flexion are happy with this, let's get it in before feature freeze on the 15th and then we'll have two weeks plus maintenance to fix any regressions or special cases if any.

flexiondotorg:
This works for me, but I'm testing on Arch and several versions ahead of Mint on Ubuntu. +1 from me.

raveit65:
+1 from me too
I use it since some weeks for fedora and now people starts whining that everything works like it should and they have to unlock the screen after suspend ;)
Which indicates that the change works brilliant.
https://bugzilla.redhat.com/show_bug.cgi?id=1271301
Maybe we should add a gsettings key to disable the lockscreen after suspend.
But this can be done later.

monsta:
> people starts whining that everything works like it should and they have to unlock the screen after suspend

That's one of the things that bothers me here... sure, it's more secure than before, but looks like it's not configurable?
Again, there were some comments on this subject in https://github.com/linuxmint/cinnamon-screensaver/pull/36, but it seems that the discussion went nowhere.

raveit65:
> That's one of the things that bothers me here... sure, it's more secure than before, but looks like it's not configurable?

People can simply disable the lookscreen . I do not really understand why people needs different settings for this. Everyone can wake up the notebook from suspend and see everthing in home directory, if you disable the lookscreen.
But as i said a missing setting for only disable the lookscreen after suspend/hybernate is not a reason to block this PR which makes it more secure.

flexiondotorg:
I agree with the comment above.

monsta:
> People can simply disable the lookscreen

Well I can't find a setting for that... I have lockscreen disabled in ```mate-screensaver-preferences``` but that setting doesn't apply here.

@clefebvre: BTW, I've noticed that second commit (https://github.com/paravoid/mate-screensaver/commit/26342dca60d131a0c55470c726a318d093338c9d) isn't present in ```cinnamon-screensaver```.

Anyway, this PR works alright overall, so I'll merge it. We can improve things later. :smile:

monsta:
Rebased and merged in https://github.com/mate-desktop/mate-screensaver/commit/306599bdbae51788ed2c2faa061adc6966fd814d and https://github.com/mate-desktop/mate-screensaver/commit/f1739c03e00a51e87bd49d04827e5b765d1c70de.

monsta:
We had to revert the second commit in https://github.com/mate-desktop/mate-screensaver/commit/372da14ddaf723f5da64413676d774223ab1b68f for now. We need this to be configurable in MATE itself, to let users decide whether they want that lockscreen or not (just like with locking on inactivity). We'll get back to improving the behavior during 1.13 dev phase.
See also https://bugzilla.redhat.com/1271301 and https://github.com/mate-desktop/mate-screensaver/issues/84.

clefebvre:
It is already.. albeit two little bugs in MSM (which only looks at one of the two settings) and MPM (which always locks no matter what), but it's in there that we should handle that logic.

PR #87, opened 2015-11-12
fix warnings due to wrong id passed to g_source_remove

monsta:


raveit65:
Where i see the warnings?
build log?
logs?

monsta:
As usual - ```~/.xsession-errors```. Looks like this:
```
GLib-CRITICAL **: Source ID 38 was not found when attempting to remove it
```

monsta:
Ah, and you'll probably need to let screensaver activate first... via doing nothing I mean :smile:
Screensaver preview in the prefs dialog didn't give any warnings for me.

raveit65:
OK, but preview or prefs gives me that warning :)
.....f...k deprecations
```
(mate-screensaver-preferences:25582): GLib-GObject-WARNING **: gsignal.c:2451: signal 'size-request' is invalid for instance '0x1ef25e0' of type 'GtkScrolledWindow'
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
```

monsta:
Well that one is different and well-known :smile:

PR #97, opened 2016-06-18
Fix black background & remove deprecations

XRevan86:


raveit65:
pop-square theme is still broken with dark themes.

raveit65:
Thank you for the great job.

PR #111, opened 2017-01-15
1.16: Avoid deprecated gdk_window_set_background functions

XRevan86:
GTK+ 3.22 has broken them completely, and in mate-screensaver there aren't really a need in them at all, in any GTK+ version.
Additionally this significantly reduces the gap between the code for GTK+2 and GTK+3.

### DIFF
--- a/data/mate-screensaver-preferences.ui
+++ b/data/mate-screensaver-preferences.ui
@@ -129,6 +129,7 @@
         <child>
           <object class="GtkDrawingArea" id="fullscreen_preview_area">
             <property name="visible">True</property>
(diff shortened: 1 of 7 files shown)


raveit65:
Just tested this with f25 (mate gtk3). The issue with blank screensaver themes is gone.
And i see no problems with other themes.

raveit65:
Same results with gtk+-2, tested with centos7.3 .

raveit65:
PS: i will use those commits to fix downstream report immediately, as reporters thinks this is a security issue.

PR #112, opened 2017-01-17
Change shake times to even number, so the dialog can return to its original position.

hao-lee:
Hi, I found a bug in `shake_dialog` function.

When the password is wrong, the dialog will shake several times to indicate that result. However, there is a bug in `shake_dialog` function.

Variable `i` will increase from `0` to `8` and this `for loop` will be executed **odd times**, so the dialog will stay on the right of its original position.

I have changed the max value of `i` to `8` and it works properly!


flexiondotorg:
Reviewd by @XRevan86 . Thanks!

PR #113, opened 2017-02-10
copy-theme-dialog: port GtkTable to GtkGrid

raveit65:
I don't know how to trigger this GUI to see it, but i used replacements for GTK_FILL and GTK_EXPAND table options.
So this might be fine.

### DIFF
--- a/src/copy-theme-dialog.c
+++ b/src/copy-theme-dialog.c
@@ -53,7 +53,7 @@ copy_theme_dialog_response (GtkDialog *dialog, gint response_id);
 static void
 eel_gtk_label_make_bold (GtkLabel *label);
 static void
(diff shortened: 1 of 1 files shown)


monsta:
It should be triggered by drag-n-dropping a ```.desktop``` file (or a bunch of such files) into preferences dialog. But I couldn't get it to appear. It just silently copies the files into ```~/.local/share/applications/screensavers``` and doesn't show any dialogs.

raveit65:
Well same here, no chance to trigger the gui.
I would merge it as it fixes deprecation build warnings and in future we need to replace it.
If someone shows us the GUI with a bug i am happy to fix it than.

monsta:
I'm fine with it. Wonder if anyone ever used this hidden feature.

PR #116, opened 2017-03-28
Switch to GdkSeat for GTK+ 3.20 and newer

XRevan86:
```gdk_keyboard_grab``` and ```gdk_pointer_grab``` are removed from GTK+ 4 making this an important change for the future.
Yet at the same time I can't be 100% confident because I'm touching (a lot) a very important part of screenlocking potentionally creating a "security issue".

### DIFF
--- a/src/gs-grab-x11.c
+++ b/src/gs-grab-x11.c
@@ -39,7 +39,7 @@
 
 static void     gs_grab_class_init (GSGrabClass *klass);
 static void     gs_grab_init       (GSGrab      *grab);
-static void     gs_grab_finalize   (GObject        *object);
+static void     gs_grab_finalize   (GObject     *object);
 
 #define GS_GRAB_GET_PRIVATE(o) (G_TYPE_INSTANCE_GET_PRIVATE ((o), GS_TYPE_GRAB, GSGrabPrivate))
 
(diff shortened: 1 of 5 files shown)


lukefromdc:
I don't have this one installed at the moment, which means I would not see any behavioral changes other than either being unable to lock or unable to unlock the screen at all.

lukefromdc:
In a quick test installation, the screensaver panel applet could not lock the screen nor activate the screensaver. The preferences dialog worked. I have no idea however if this is ongoing or a result of this PR. Build was fine.

XRevan86:
@lukefromdc, maybe you should have started the mate-screensaver daemon.

lukefromdc:
Starting mate-screensaver, then using the screensaver applet led to the screensaver being displayed for a few seconds, then crashing and returning to the desktop. Same if a new session was started with the screensaver daemon active. This really needs testing by someone who normally uses it and has it installed though,

flexiondotorg:
Erm, did we establish if this actually works?

XRevan86:
@flexiondotorg, this most definitely basically works %).
Testing is still desired. Some freaky ways of trying to break the grab of mate-screensaver on the keyboard or something.

sbalneav:
I meant to approve the merge, instead I just did it.  But I have been testing it all day, and it appears to work just fine for me.

flexiondotorg:
OK, good to hear ☺️

raveit65:
I see no regression here.
Restarting the daemon after update and using the lock applet works out of box here.
In a VM i restarted the session and  locked the screen several times, no problems.

chr-chr:
Hello,

this patch plus another bug makes this a security issue.

https://github.com/mate-desktop/mate-screensaver/issues/180

raveit65:
No

chr-chr:
> No

You're right. That code was buggy beforehand.

PR #117, opened 2017-03-28
Implement OpenBSD authentication using bsd_auth(3).

rnagy:
OpenBSD does not have nor use pam(8) for password authentication but
instead uses bsd_auth(3): add a bsd_auth authentication scheme to
make locking possible.
This has been a part of gnome-screensaver already but was probably
removed at one point. This is mostly identical to that code, with some
minor modifications.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -532,6 +532,44 @@ if test "x$have_libgl" = "xyes"; then
 fi
 
 dnl ---------------------------------------------------------------------------
+dnl - Check for bsd_auth(3) (OpenBSD)
+dnl ---------------------------------------------------------------------------
+have_bsdauth=no
+with_bsdauth_req=unspecified
+NEED_SETUID=no
+
+case "$host" in
+  *-openbsd*)
+    with_bsdauth=yes
+    AUTH_SCHEME=bsdauth
+    NEED_SETUID=no
+    if test "x$enable_locking" = "xyes"; then
+      with_bsdauth_req=yes
+      NEED_SETUID=yes
+    fi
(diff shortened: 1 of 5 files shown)


raveit65:
I have no idea how we can test this without using OpenBSD? 
Looks like this change does not affect other distros if you don't enable it in configure.
@XRevan86 @monsta @sbalneav 
Guys can you please do a code review?

monsta:
> This has been a part of gnome-screensaver already but was probably
removed at one point. 

Actually, it was only added in 3.2.1 (see [here](https://git.gnome.org/browse/gnome-screensaver/commit/?id=d7ab99c00a01bea870cf51d855ecf320e0d23bac) and [here](https://git.gnome.org/browse/gnome-screensaver/commit/?id=169c76373454b8b4e2d25b27f8d80d3494845cb3)), and MATE was forked from 2.30, that's why we don't have it... It's a good addition, only that we can't test it as we only use Linux.

Can bsd_auth be used on NetBSD? If yes, we can ask @obache to test it.

monsta:
The code seems ok to me, but I wonder if we should add the new source file to "extra dist" or something like that. Need to test ```make distcheck``` with this to see if it breaks or not.

obache:
@monsta BSD Auth is only used on OpenBSD nowadays.

rnagy:
@monsta BSD auth is OpenBSD only and i've been running with this patch for weeks now without any issues, works as expected.

monsta:
Ok, merged and added the fix for the tarball.

PR #118, opened 2017-04-10
lock-dialog UI: avoid deprecated widgets

raveit65:
Please test.

### DIFF
--- a/data/lock-dialog-default.ui
+++ b/data/lock-dialog-default.ui
@@ -8,335 +8,33 @@
     <property name="label_xalign">0</property>
     <property name="shadow_type">out</property>
     <child>
-      <object class="GtkAlignment" id="alignment1">
+      <object class="GtkBox" id="lock-box">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
-        <property name="border_width">12</property>
+        <property name="border_width">30</property>
+        <property name="orientation">vertical</property>
         <child>
-          <object class="GtkBox" id="lock-box">
+          <object class="GtkNotebook" id="notebook">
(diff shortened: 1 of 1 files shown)


monsta:
I've used a VM so I could take the screenshots. It has GTK+ 3.14 though... Anyway, the logout dialog border got thinner after this PR:

virtualbox_lmde 2_14_04_2017_14_51_50

virtualbox_lmde 2_14_04_2017_14_59_06

Should we keep this change? I don't know which one is better.

raveit65:
Well i saw that but i wasn't really sure if there was a different as i did that on baremetal, it should be possible to set a margin on the GtkBox.
It looks a bit more compact now, which isn't bad imo.
If you really disagree i will spend some more time ;-)

More from interest for me is why we have the annoying black frame behind the window, which makes it impossible to increase the border-radius in themes.
Because we don't use a compositor here?
If yes, why we don't use WM settings from session?

raveit65:
@monsta 
increase the border-size at line 14 to 30 (12+18=30) 
```
-      <object class="GtkAlignment" id="alignment1">
+      <object class="GtkBox" id="lock-box">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
-        <property name="border_width">12</property>
+        <property name="border_width">18</property>
+        <property name="orientation">vertical</property>
```
But , it's really hard to say :-)
I will do what you like.....

raveit65:
Rebased and using 30px border-width for lock-box.

monsta:
I didn't really disagree, I just couldn't decide which one is better, old or new :slightly_smiling_face:

PR #119, opened 2017-04-12
Bug Fix: g_get_real_name will never return NULL pointer

hao-lee:
The function `g_get_real_name` is used to get user's full name, but when the full
name is not existed, it will return an empty string instead of `NULL`.

### DIFF
--- a/src/gs-lock-plug.c
+++ b/src/gs-lock-plug.c
@@ -1595,7 +1595,8 @@ get_user_display_name (void)
 
 	name = g_get_real_name ();
 
-	if (name == NULL || strcmp (name, "Unknown") == 0)
+	if (name == NULL || g_strcmp0 (name, "") == 0 ||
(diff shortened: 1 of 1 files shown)


XRevan86:
I'd do it like this:
```diff
-	if (name == NULL || strcmp (name, "Unknown") == 0)
+	if (name == NULL || g_strcmp0 (name, "") == 0 ||
+           g_strcmp0 (name, "Unknown") == 0)
```

hao-lee:
@XRevan86 Yeah, this way could handle all possible results. :smile:

raveit65:
@hao-lee 
Can you please update PR ?

raveit65:
merged git merge --ff-only hao-lee-master
Thank you

hao-lee:
All right. I fetch the latest commits and find this line has been changed....The reaction is so fast...

PR #153, opened 2018-04-26
Support new mate-menus API

yetist:
Before that, I submitted a [PR](https://github.com/mate-desktop/mate-menus/pull/59) for mate-menus, if that PR merged, it will break mate-screensaver.

So now I've changed the mate-screensaver code to use the new api of mate-menus 1.21.0.


raveit65:
Which part of gui needs to tested?
Or where can i see changes?
Or does everything works in the background?

monsta:
Do we accept version bumps here, or do we add these ourselves after merging?

src/gs-theme-manager.c
```diff
@@ -397,12 +378,19 @@ static MateMenuTree *
...
 {
 	MateMenuTree *themes_tree = NULL;
+	GError *error = NULL;
 
 	/* we only need to add the locations to the path once
 	   and since this is only run once we'll do it here */
 	add_known_engine_locations_to_path ();
 
-	themes_tree = matemenu_tree_lookup ("mate-screensavers.menu", MATEMENU_TREE_FLAGS_NONE);
+	themes_tree = matemenu_tree_new ("mate-screensavers.menu", MATEMENU_TREE_FLAGS_SORT_DISPLAY_NAME);
```
monsta: Seems like this sorting isn't used anywhere in the code, so probably there's no need to add it.

raveit65:
Version bumping or editing NEWS files should be done from release team in general, as it depends on other things. And i see no package which needs to be rebuild because mate-screensaver.
@yetist 
Can you please revert this?

raveit65:
Same in other PRs, we all know that we have to bump build requires to match new mate-menus.

src/gs-theme-manager.c
yetist: The screensaver was not sorted before, and it's ok now. Don't you like it?

monsta: Maybe I missed the sorting - I didn't notice any differences. Where should I look?

monsta:
Ok, version bump is left in mate-menus PR. If it's removed, then the build requirement bump here has to go as well.

monsta:
Looks ok, I'm just a bit confused abort sorting (see my comments above).

yetist:
Sorted screensaver themes here:

monsta:
@raveit65: the last commit reverts the menu sorting on loading, which doesn't seem to affect anything (the screensaver list is sorted somewhere else apparently).

raveit65:
Squashed all in one commit and merged.
https://github.com/mate-desktop/mate-screensaver/commit/c3807e72fc761778100f500c0966fe491aec7fd2
Thanks for your work.

raveit65:
mate-screensaver-1.21 is released but not tagged.
Tarball is at pre-release folder for testing.

PR #165, opened 2018-08-13
lock-plug: Only scale lock-screen widgets on older versions of GTK

vkareh:
This fixes an issue where the lock screen widgets look wrong on GTK+ >= 3.23


raveit65:
This fixes the issue with the lockscreen-dialog and gtk+-3.23.1
Before:
lockscreen
After:
lockscreen-fine
Thank you

raveit65:
i will cherry-pick that to 1.20 branch

PR #166, opened 2018-08-14
gs-manager.c: add missing debug information

zhangxianwei8:
Print monitor in right place:
https://github.com/mate-desktop/mate-screensaver/blob/ef8f73d6f81ac3d841bf136960adf523204412ea/src/gs-manager.c#L1542-L1543
Please review and test:) @sc0w @raveit65

### DIFF
--- a/src/gs-manager.c
+++ b/src/gs-manager.c
@@ -1539,8 +1539,8 @@ gs_manager_create_window_for_monitor (GSManager  *manager,
 	display = gdk_monitor_get_display (monitor);
 	gdk_monitor_get_geometry (monitor, &rect);
 
-	gs_debug ("Creating a window for the monitor [%d,%d] (%dx%d)",
-	          monitor, rect.x, rect.y, rect.width, rect.height);
+	gs_debug ("Creating a window [%d,%d] (%dx%d)",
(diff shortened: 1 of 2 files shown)


monsta:
Well... the `monitor` argument is a pointer to `GdkMonitor`, not an integer.
It was changed in https://github.com/mate-desktop/mate-screensaver/commit/5d4416aa5853de486e62ca97207ee1da4b35b199#diff-d417e0cf467627785b3a651a7990c54eR1546 for GTK+ 3.22, and later we dropped the code for older GTK+ versions.

zhangxianwei8:
That means the `monitor` shoud be removed? If so, this code should be fixed too:
https://github.com/mate-desktop/mate-screensaver/blob/ef8f73d6f81ac3d841bf136960adf523204412ea/src/gs-window-x11.c#L390-L395

monsta:
Yeah, the variable in the second debug message is a `GdkMonitor` too. If there's no way to get the integer number of the monitor now, then it's ok to drop these parts from the debug message. Maybe also drop "for the monitor"/"on monitor" text too, as it's not useful without the monitor number.

zhangxianwei8:
Updated the PR.

PR #169, opened 2018-09-08
fade: fix fade computation math

raveit65:
commit https://github.com/mate-desktop/mate-screensaver/commit/599bc81
changed the fade time out to be less than a 1000ms.  Unfortunately, this broke the fade because the
math for computing the number of fade steps used integer arithmetic,
and performed a division by a 1000.

https://bugzilla.gnome.org/show_bug.cgi?id=672583

origin commit:
https://gitlab.gnome.org/GNOME/gnome-screensaver/commit/7e2085a

### DIFF
--- a/src/gs-fade.c
+++ b/src/gs-fade.c
@@ -755,7 +755,7 @@ gs_fade_start (GSFade *fade,
 	{
 		guint num_steps;
 
-		num_steps = (fade->priv->timeout / 1000) * steps_per_sec;
+		num_steps = (fade->priv->timeout / 1000.0) * steps_per_sec;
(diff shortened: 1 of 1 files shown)


XRevan86:
Makes sense.

vkareh:
Agreed, I think we should just merge this. :)

vkareh:
What? No! I did rebase and merge! Why did it create a new commit!?!?

raveit65:
github UI is sometimes weird.
Personal i prefer command line.
git rebase master && git checkout master && git merge --ff-only the-branch-name

PR #178, opened 2019-01-22
Prompt-driven auth. helper

wolneykien:
Hi, MATE developers! I want to announce the prompt-driven authentication helper. By the term "prompt-driven" I mean two-way conversation between the screensaver dialog and the PAM stack. As you probably know, PAM works by conversation with a dialog program asking the user to enter something in answer to each message it sends. In the most conventional case, the only question is "Password:" and the password is the only data the user enters. But in general, the number of questions and messages are not limited to that.

The previous (i.e. current) version of PAM helper was written and works for the mentioned "only password" authentication scheme. For other schemes it's not enough. The implementation I'm announcing here fixes that limitation.


wolneykien:
Hi. I'm talking about the PAM **helper** binary, not the PAM itself. The PAM itself stays the same for years, but the helper doesn't use most of PAM's features. The version I announce does.

yetist:
I saw that the API you added could read and write messages and prompts, but did not see the actual use, so I wanted to know how to test it.

raveit65:
> Hi. I'm talking about the PAM helper binary, not the PAM itself.

Ok, please keep in mind that most of us aren't specific maintainers of one repo and we work on all MATE repos. Sorry, i don't know very much about pam helper specs.
So giving more infos and steps to test (before/after) would help a lot to review this especially this looks like a huge change.
Thanks.

wolneykien:
> I saw that the API you added could read and write messages and prompts, but did not see the actual use, so I wanted to know how to test it.

The use is in the `src/gs-auth-helper.c`, see https://github.com/mate-desktop/mate-screensaver/pull/178/commits/c2e2eae696431ed6265b8b303192f7172743ec49 for example. In order to make a test-run the `mate-screensaver` should be built with `--enable-authentication-scheme=helper` parameter passed to its `./configure` script. Then you just lock the screen and perform the unlock procedure. Under the hood it should start `mate-screensaver-pam-helper` binary and use it to communicate with PAM.

<hr />

> Ok, please keep in mind that most of us aren't specific maintainers of one repo…

Okay

raveit65:
For better understanding. Without `--enable-authentication-scheme=helper` the lockscreen use pam directly?
I never used this option to build mate-screensaver for fedora.
https://src.fedoraproject.org/cgit/rpms/mate-screensaver.git/tree/mate-screensaver.spec#n86

src/gs-auth-helper.c
```diff
@@ -78,13 +78,14 @@ gs_auth_get_verbose (void)
 
 static gboolean
 ext_run (const char *user,
-         const char *typed_passwd,
-         gboolean    verbose)
+		 GSAuthMessageFunc func,
```
raveit65: Indents, please do not use tabs if possible.

wolneykien:
> For better understanding. Without `--enable-authentication-scheme=helper` the lockscreen use pam directly?

Probably, yes. There are the following lines in `configure.ac`:

    elif test x$enable_authentication_scheme = xauto ; then
	if test x$have_pam != xno ; then
		AUTH_SCHEME="pam"

And the default is `"auto"`.

> I never used this option to build mate-screensaver for fedora.
> https://src.fedoraproject.org/cgit/rpms/mate-screensaver.git/tree/mate-screensaver.spec#n86

That's interesting. If the local (i.e. `/etc/passwd`, `/etc/shadow`, `/etc/tcb`) authentication is accessible without root privileges on Fedora? I'm asking, because the screensaver process is started from a normal, unprivileged user session.

wolneykien:
> Is it possible to squash commits, where your fix previous commits, into previous commits?
> To reduce the count of commits?

Of course.

raveit65:
> That's interesting. If the local (i.e. /etc/passwd, /etc/shadow, /etc/tcb) authentication is accessible without root privileges on Fedora? I'm asking, because the screensaver process is started from a normal, unprivileged user session.

Not sure, but i think all this will be handle by systemd-logind in fedora or distros with systemd.
And m-s call systemd-logind in `src/gs-listener-dbus.c` ,......... former console-kit.

raveit65:
Only /etc/passwd can be read by normal user.
```
[rave@mother ~]$ cat /etc/shadow
cat: /etc/shadow: Permission denied
[rave@mother ~]$ cat /etc/tcb
cat: /etc/tcb: No such file or directory
```

lukefromdc:
Any progress on this or should we close it?

wolneykien:
Ok, I'll just rebase these now!

wolneykien:
I've fixed the `Makefile.am` conflict and squash most of the WIP fixes into one commit. There is no conflicts now, but the next question is: does it work now?

wolneykien:
I see that the Travis builds fail now. However, the code is known to work in ALT (https://packages.altlinux.org/en/sisyphus/srpms/mate-screensaver). And that's how it is configured and built: https://github.com/wolneykien/mate-screensaver/blob/sisyphus/mate-screensaver.spec#L37 .

raveit65:
Do we need a new dependency?
Looks like there is something missing.
https://travis-ci.org/github/mate-desktop/mate-screensaver/jobs/661878392#L2644
```
/usr/bin/ld: gs-auth-pam.o: in function `pam_conversation':

gs-auth-pam.c:(.text+0x314): undefined reference to `pam_style_to_gs_style'

collect2: error: ld returned 1 exit status

make[3]: *** [Makefile:905: mate-screensaver-dialog] Error 1
```

raveit65:
Build is with `./configure --enable-compile-warnings=maximum`
https://travis-ci.org/github/mate-desktop/mate-screensaver/jobs/661878392#L1094
That means
a) build is without your new code 
```
Have password helper:     no
```
b) and build failed without `password helper` code, which should be fixed, or not?

In a second step missing config option can be add to travis config to test building your code.

wolneykien:
Just built with `./configure --enable-compile-warnings=maximum` locally, on my machine:

    Have password helper:     no
    Authentication scheme:    pam

And  `src/gs-auth-pam.o` builds OK without the ld error above. The `pam_style_to_gs_style()` function that is missing in the Travis build is actually _inline_ and is defined in `src/gs-auth-pam.h`: https://github.com/wolneykien/mate-screensaver/blob/announce/src/gs-auth-pam.h#L29 (and the latter is included in `src/gs-auth-pam.c` here: https://github.com/wolneykien/mate-screensaver/blob/announce/src/gs-auth-pam.c#L48).

raveit65:
I get same build failure when trying to build local from your branch (fedora 30).
```
  CCLD     mate-screensaver-dialog
/usr/bin/ld: gs-auth-pam.o: in function `pam_conversation':
gs-auth-pam.c:(.text+0x314): undefined reference to `pam_style_to_gs_style'
collect2: error: ld returned 1 exit status
make[3]: *** [Makefile:905: mate-screensaver-dialog] Error 1
```

wolneykien:
But `pam_style_to_gs_style()` is actually defined, isn't it? What do you think?
There are, however, `G_BEGIN_DECLS` / `G_END_DECLS` parentheses around it. I'll investigate that tonight.

wolneykien:
Hi! I think I've fixed the previous error by adding `static` to `inline` in the definition of `pam_style_to_gs_style()`. It now compiles successfully on Travis, but the next stage (`/rootdir/after_scripts`) fails with strange diagnostics: `gs-auth-pam.h not found`. I don't fully understand your build scripts, but if `gs-auth-pam.h` was found and used during normal compilation then why was it absent afterwards? The file is anyway present in the git tree.

helper/Makefile.am
```diff
@@ -0,0 +1,26 @@
...
+	$(MATE_SCREENSAVER_CFLAGS)
+
+mate_screensaver_pam_helper_SOURCES =	\
+	pam-helper.c
+
+mate_screensaver_pam_helper_LDADD =	\
+	libhelper-proto.a
+
+mate_screensaver_pam_helper_LDFLAGS =	\
+	$(AUTH_LIBS)
```
rbuj, line 26: You may check the helper name in configure.ac file and use a conditional to enable its building:
ENABLE_PASSWD_HELPER + ! HAVE_PASSWD_HELPER
```
--with-passwd-helper=mate-screensaver-pam-helper
```
or it has been enabled and the path has not been provided. Therefore, I think you need to check whether is enabled and the name of the helper.

configure.ac
```diff
@@ -1118,6 +1118,7 @@ data/images/cosmos/Makefile
 savers/Makefile
 doc/Makefile
 doc/mate-screensaver.xml
+helper/Makefile
```
rbuj, line 1121: the creation of this file is conditional

Makefile.am
```diff
@@ -9,7 +9,12 @@ DISTCHECK_CONFIGURE_FLAGS = \
 	--without-systemd \
 	CFLAGS='-Wno-deprecated-declarations'
 
-SUBDIRS =			\
+SUBDIRS =
+if HAVE_PASSWD_HELPER
+SUBDIRS += helper
+endif
```
rbuj, line 15: you don't need to initialize SUBDIRS with empty

src/gs-auth-pam.c
```diff
@@ -45,6 +45,7 @@
 #include <gtk/gtk.h>
 
 #include "gs-auth.h"
+#include "gs-auth-pam.h"
```
rbuj, line 48: It does not make sense placing the file gs-auth-pam.h in the src folder.

src/gs-auth-pam.h
```diff
@@ -0,0 +1,54 @@
...
+		break;
+	case PAM_ERROR_MSG:
+		style = GS_AUTH_MESSAGE_ERROR_MSG;
+		break;
+	default /* PAM_TEXT_INFO */:
+		style = GS_AUTH_MESSAGE_TEXT_INFO;
+	}
+
+	return style;
+}
```
rbuj, line 50: You can also put the code out of the h file.

helper/pam-helper.c
```diff
@@ -0,0 +1,289 @@
...
+
+	if (!(reply = malloc(sizeof(*reply) * num_msg)))
+		return PAM_CONV_ERR;
+
+	for (num = 0; num < num_msg; num++) {
+		ssize_t wt, rd;
+		size_t msg_len = strlen(msg[num]->msg);
+		wt = write_prompt (STDOUT_FILENO,
+						   pam_style_to_gs_style (msg[num]->msg_style),
+						   msg[num]->msg, msg_len);
```
rbuj, line 107: fix indent

helper/pam-helper.c
```diff
@@ -0,0 +1,289 @@
...
+	sigaction(SIGALRM, &action, NULL);
+}
+
+static int
+_converse(int num_msg, const struct pam_message **msg,
+		struct pam_response **resp, void *appdata_ptr)
+{
+	struct	pam_response *reply;
+	char buf[MAXLEN];
+	int	num;
```
rbuj, line 96: don't use tab between variable type and name

helper/pam-helper.c
```diff
@@ -0,0 +1,289 @@
...
+	sigaction(SIGINT, &action, NULL);
+	sigaction(SIGQUIT, &action, NULL);
+	sigaction(SIGALRM, &action, NULL);
+}
+
+static int
+_converse(int num_msg, const struct pam_message **msg,
+		struct pam_response **resp, void *appdata_ptr)
+{
+	struct	pam_response *reply;
```
rbuj, line 94: don't use tab between variable type and name

src/gs-auth-pam.c
wolneykien: Hm... What about all other `.h` files in the `src/` folder?

helper/pam-helper.c
wolneykien: Thanks! Did you use any parser to find that?

rbuj: no

src/gs-auth-pam.c
rbuj: I didn't see any other relative path in src folder.
```
#include "../src/gs-auth-pam.h"
```

rbuj: > Moving it outside  of `src/` would simply require an other relative path (or `-I` CC flag).
+1

src/gs-auth-pam.h
wolneykien: I'll try to include it in the `libhelper-proto.a` as a normal, non-inline function.

src/gs-auth-pam.c
rbuj: > Anyway, I don't see how the last Travis error is related to this.

Using top_srcdir or similar variable may help.
https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Preset-Output-Variables.html

wolneykien: Hm... But here we are discussing why `gs-auth-pam.h` is placed in `src/`. There are a number of other `.h` files in `src/`. One example is `gs-auth.h`, another one is `subprocs.h`. Both are included in `gs-auth-pam.c` without errors. However, the compiler is unable to find `gs-auth-pam.h` that is placed next to `gs-auth.h` in the same folder.

One more thought. I've just noticed the relative path in the compilation error message:
```
../../../src/gs-auth-pam.c:48:10: fatal error: gs-auth-pam.h: No such file or directory
```
That path (`../../../src`) looks strange because then it reports that it is leaving `src/`:
```
make[4]: Leaving directory '/rootdir/mate-screensaver-1.25.0/_build/sub/src'
```
Thus, I think that's **another** `src/` where the compilation fails. Probably, that isn't a normal compilation and is related to some special part in the Travis build scripts.

rbuj: Did you try to not use relative paths?
```diff
diff --git a/helper/Makefile.am b/helper/Makefile.am
index 205a290..3907467 100644
--- a/helper/Makefile.am
+++ b/helper/Makefile.am
@@ -14,7 +14,8 @@ pkglibexecdir = $(libexecdir)/mate-screensaver
 pkglibexec_PROGRAMS = mate-screensaver-pam-helper
 
 mate_screensaver_pam_helper_CFLAGS =   \
-       $(MATE_SCREENSAVER_CFLAGS)
+       $(MATE_SCREENSAVER_CFLAGS)      \
+       -I$(top_srcdir)/src
 
 mate_screensaver_pam_helper_SOURCES =  \
        pam-helper.c
diff --git a/helper/pam-helper.c b/helper/pam-helper.c
index 2c8c535..1ab8792 100644
--- a/helper/pam-helper.c
+++ b/helper/pam-helper.c
@@ -28,7 +28,7 @@
 #include <ctype.h>
 
 #include "helper_proto.h"
-#include "../src/gs-auth-pam.h"
+#include <gs-auth-pam.h>
 
 #define MAXLEN 1024
 
```

wolneykien: Ok, I'll try. But the reported problem is not in `helper/pam-helper.c` but in `src/gs-auth-pam.c`.

Makefile.am
wolneykien: That's not true:
```
$ autoreconf -fisv
...
Makefile.am:13: error: SUBDIRS must be set with '=' before using '+='
Makefile.am:15: error: cannot apply '+=' because 'SUBDIRS' is not defined in
Makefile.am:15: the following conditions:
Makefile.am:15:   !HAVE_PASSWD_HELPER
Makefile.am:15: either define 'SUBDIRS' in these conditions, or use
Makefile.am:15: '+=' in the same conditions as the definitions.
```

rbuj:
```diff
diff --git a/Makefile.am b/Makefile.am
index f723689..1b9ea3f 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -9,12 +9,12 @@ DISTCHECK_CONFIGURE_FLAGS = \
        --without-systemd \
        CFLAGS='-Wno-deprecated-declarations'
 
-SUBDIRS =
 if HAVE_PASSWD_HELPER
-SUBDIRS += helper
+HELPER_DIR = helper
 endif
 
-SUBDIRS +=     \
+SUBDIRS =                      \
+       $(HELPER_DIR)           \
        po                      \
        src                     \
        savers                  \
```
How do you define HAVE_PASSWD_HELPER?
I hope you don't do it that way:
```
$ ./autogen.sh --prefix=/usr --with-passwd-helper=/path/helper && make V=1 &> make.log && grep helper make.log
```

helper/Makefile.am
```diff
@@ -4,7 +4,7 @@ AUTOMAKE_OPTIONS = 1.7
 noinst_LIBRARIES = libhelper-proto.a
 
 libhelper_proto_a_CFLAGS =	\
-	$(MATE_SCREENSAVER_CFLAGS)
```
rbuj: There must be only one argument on each line, use the backslash to escape the end of the assignment.

helper/Makefile.am
```diff
@@ -14,7 +14,7 @@ pkglibexecdir = $(libexecdir)/mate-screensaver
 pkglibexec_PROGRAMS = mate-screensaver-pam-helper
 
 mate_screensaver_pam_helper_CFLAGS =	\
-	$(MATE_SCREENSAVER_CFLAGS)
```
rbuj: same here

Makefile.am
wolneykien: > How do you define HAVE_PASSWD_HELPER?
> I hope you don't do it that way:
> 
> ```
> $ ./autogen.sh --prefix=/usr --with-passwd-helper=/path/helper && make V=1 &> make.log && grep helper make.log
> ```

What's wrong with it?

rbuj: The path set by using --with-passwd-helper may not be the same of the installed binary.

wolneykien: Yes, but the path to the helper **binary** isn't used at build time.

I think, I've found the solution to fix the Travis builds. Finally I've noticed, that the error is in `make distcheck`. And in order to fix it the _new_ file `gs-auth-pam.h` should be included into the distribution --- i.e. listed as a source in `Makefile.am`.

rbuj: > Yes, but the path to the helper **binary** isn't used at build time.

Are you sure?
```
$ ./autogen.sh --prefix=/usr --with-passwd-helper=/path/to/helper
$ grep -R PASSWD_HELPER_PROGRAM --include=*.c --include=config.h --include=configure.ac
src/gs-auth-helper.c:		           PASSWD_HELPER_PROGRAM, user);
src/gs-auth-helper.c:		execlp (PASSWD_HELPER_PROGRAM, PASSWD_HELPER_PROGRAM, "mate-screensaver", user, NULL);
src/gs-auth-helper.c:			g_message ("%s: %s", PASSWD_HELPER_PROGRAM, g_strerror (errno));
src/gs-auth-helper.c:	if (g_access (PASSWD_HELPER_PROGRAM, X_OK) < 0)
src/gs-auth-helper.c:		           PASSWD_HELPER_PROGRAM);
configure.ac:    AC_DEFINE_UNQUOTED(PASSWD_HELPER_PROGRAM, "$with_passwd_helper", [Full pathname of password helper application])
config.h:#define PASSWD_HELPER_PROGRAM "/path/to/helper"
```

wolneykien: Hm... `--with-passwd-helper` option defines `PASSWD_HELPER_PROGRAM` variable that is substituted in `src/gs-auth-helper.c`. The path is simply hardcoded into the object file and can not be changed later (I think, by means of security). However, the resulting `src/gs-auth-helper.o` is used for _two purposes_: as a part of the main screensaver binary that _is not_ used at compile time, and in the test program `src/test-passwd` that probably _can be used_ in some auto-tests. So, potentially, we have a problem with `test-password`, while having the hardcoded path to the real (installed) helper in the main program seems to be ok.

rbuj: How do you call mate-screensaver-pam-helper?
```
$ grep -R mate-screensaver-pam-helper --include=*.h --include=*.c
```

helper/pam-helper.c
```diff
@@ -0,0 +1,289 @@
...
+	UNIX_FAILED = 1
+};
+
+static char * program_name;
+
+/*
+ * Log error messages
+ */
+static void
+_log_err(int err, const char *format,...)
```
rbuj, line 46: IMHO, you don't need to use vsyslog directly, you can set G_LOG_DOMAIN and use GLIB's message output functions.

src/gs-auth-helper.c
```diff
@@ -99,18 +103,23 @@ ext_run (const char *user,
 
 	if ((pid = fork ()) < 0)
 	{
-		close (pfd [0]);
-		close (pfd [1]);
+		close (pfd [0]); close (pfd [1]);
+		close (r_pfd [0]); close (r_pfd [1]);
```
rbuj, line 107: one sentence per line

rbuj: Why do you need 4 pipes?

helper/pam-helper.c
wolneykien: The `pam-helper` is very simple, it isn't linked with GLib.

rbuj: why MATE_SCREENSAVER_CFLAGS?

src/gs-auth-helper.c
wolneykien: 2 pipes, 4 descriptors actually.

> `pipefd[0]` refers to the read end of the pipe. `pipefd[1]` refers to the write end of  the pipe.

(pipe(2))

Both processes are doing read and write operations, so 4 paired descriptors are needed.

rbuj: Yes, I apologize. To make the review task more agile, you can add more information about the design patterns you used, such as the communication type (synchronous/asynchronous, semaphore/mutex, ...) in the description.

wolneykien: Updated.

helper/pam-helper.c
wolneykien: Fixed. Now `PAM_HELPER_CFLAGS`.

src/gs-auth-pam.h
wolneykien: Fixed. That was an error with `make distcheck`.

src/gs-auth-pam.c
wolneykien: Added `gs-auth-pam.h` to `AUTH_SOURCES`.

configure.ac
wolneykien: It _should_ be conditional you mean? Ok, I'll do that.

wolneykien: Later I've thought that's nothing wrong to generate the `helper/Makefile` unconditionally: the real question is about the conditions when to _build_ it. So I've made some modifications to processing of the `--enable-authentication-scheme` and `--with-password-helper` options. Now the builtin helper from `helper/` is built only when `--enable-authentication-scheme=helper` but no `--with-password-helper=...` is specified.

raveit65:
OMG.
This branch has 21 commits.
```
[rave@mother mate-screensaver]$ git format-patch dfad698c6914d376f899c629c5805377c1766ec1
0001-Read-messages-from-the-helper-draft.patch
0002-Added-helper-protocol-functions.patch
0003-Use-the-helper-protocol-to-read-the-prompt-and-write.patch
0004-Build-and-use-the-helper-library.patch
0005-Fixed-read_prompt-invocation.patch
0006-Use-helper-protocol-in-the-helper-itself.patch
0007-Add-AUTH_LIBS-lpam-when-linking-the-helper.patch
0008-Add-helper-Makefile-to-the-set-of-generated-files.patch
0009-Various-fixes-to-make-the-new-code-work.patch
0010-Fixed-inlining-of-pam_style_to_gs_style-use-static-i.patch
0011-Fixed-indentation-while-assume-the-tab-width-is-8.patch
0012-Fixed-spaces-between-types-and-variables.patch
0013-Use-I-top_srcdir-src-instead-of-the-related-include-.patch
0014-Fix-List-gs-auth-pam.h-as-a-source-file-to-fix-make-.patch
0015-Fixed-CFLAGS-syntax-one-directive-per-line.patch
0016-Updated-the-copyright-message-and-the-brief-descript.patch
0017-Updated-the-default-name-of-the-helper.patch
0018-Fix-close-file-descriptors-one-per-line.patch
0019-Calculate-and-user-PAM_HELPER_CFLAGS-for-glib-2.0-he.patch
0020-Distinct-between-the-builtin-or-an-external-password.patch
0021-Remove-unused-with_passwd_helper_req-var.patch
```
Can we rebase all to one commit or do you want reduce commits for yourself?

wolneykien:
> OMG.
> This branch has 21 commits.
> ...
> Can we rebase all to one commit or do you want reduce commits for yourself?

Of course. Single commit is ok?

helper/pam-helper.c
```diff
@@ -1,15 +1,16 @@
...
- * pam_sm_authenticate function if the function notices
- * that it's unable to get the password from the shadow file
- * because it doesn't have sufficient permissions.
+ * It is supposed to be called from mate-screensaver
+ * in order to communicate with Linux PAM as a privileged proxy.
+ * The conversation messages from the PAM stack is transmitted to
+ * mate-screensaver dialog via stdout and the received user replies
+ * read from stdin are sent back to PAM.
  *
- * Copyright (C) 2002 SuSE Linux AG
```
raveit65: Why removing old copyright holder OpenSuse?
I think it is OK when you add your name to Copyright . But don't remove previous author.
Some times i think software developer act like sprayers who tagged their name on every wall ;)

raveit65:
Does it makes sense to merge all in one monster commit?
Maybe it is possible to squash  some commits? (The fix of a commit into the origin commit)
Personal, i am editing commits before i add another one to fix my previous commit.
Are you familiar with the `git rebase -i <commit_hash_before_your_PR> ?
But before you break you branch we can merge it like it is , not nice but.....
But Copyright fom OpenSuse shouldn't be removed.

src/gs-auth-helper.c
```diff
@@ -103,8 +103,10 @@ ext_run (const char *user,
 
 	if ((pid = fork ()) < 0)
 	{
-		close (pfd [0]); close (pfd [1]);
-		close (r_pfd [0]); close (r_pfd [1]);
+		close (pfd [0]);
+                close (pfd [1]);
+		close (r_pfd [0]);
+                close (r_pfd [1]);
 		return FALSE;
```
raveit65, line 110: Mixed usage of tabs and spaces.

src/gs-auth-helper.c
raveit65: Ok, you ignoring me.
You added tabs to the file

raveit65: Our code style is not to use tabs.

raveit65: Why?
```
 static gboolean
 ext_run (const char *user,
-         const char *typed_passwd,
-         gboolean    verbose)
+		 GSAuthMessageFunc func,
+		 gpointer   data)
```

raveit65:
I squashed locally all commits in one, i think it is OK.
But it has code-style issues.

wolneykien:
> I squashed locally all commits in one, i think it is OK.
> But it has code-style issues.

Tabs and OpenSuse copyright?

src/gs-auth-helper.c
raveit65: 1. format issue

wolneykien:
Single commit (with annotation), untabified all my changes (tab width is 8), OpenSuse copyright is kept.

src/gs-auth-helper.c
```diff
@@ -78,15 +81,16 @@ gs_auth_get_verbose (void)
...
 ext_run (const char *user,
-         const char *typed_passwd,
-         gboolean    verbose)
+         GSAuthMessageFunc func,
+         gpointer   data)
 {
-	int   pfd[2], status;
+        int pfd[2], r_pfd[2], status;
 	pid_t pid;
+        gboolean verbose = gs_auth_get_verbose ();
```
raveit65, line 90: Indent of Line 88 doesn' t match the other lines . 
No idea, why you don't see that before you update PR.

src/gs-auth-helper.c
```diff
@@ -99,18 +103,25 @@ ext_run (const char *user,
...
+                close (r_pfd [0]);
+                close (r_pfd [1]);
 		return FALSE;
 	}
 
 	if (pid == 0)
 	{
 		close (pfd [1]);
+                close (r_pfd [0]);
 		if (pfd [0] != 0)
```
raveit65, line 117: Indents

src/gs-auth-helper.c
```diff
@@ -99,18 +103,25 @@ ext_run (const char *user,
...
 		{
 			dup2 (pfd [0], 0);
 		}
+                if (r_pfd [1] != 1)
+                {
+                        dup2 (r_pfd [1], 1);
+                }
```
raveit65, line 125: Indents Line 121 -124

src/gs-auth-helper.c
```diff
@@ -123,40 +134,52 @@ ext_run (const char *user,
...
-			           g_strerror (errno));
-		}
-
-		unblock_sigchld ();
-		return FALSE;
 	}
 
+        close (pfd [1]);
+        close (r_pfd [0]);
 	unblock_sigchld ();
```
raveit65, line 172: Indents

src/gs-auth-helper.c
```diff
@@ -123,40 +134,52 @@ ext_run (const char *user,
...
 
 	if (! WIFEXITED (status) || WEXITSTATUS (status) != 0)
 	{
-		return FALSE;
+                ret = FALSE;
 	}
+        else
+                ret = TRUE;
```
raveit65, line 180: Indents

raveit65:
I don't get why you don't see all this code-style issues for yourself?
No tabs no pain ;)

src/gs-auth-helper.c
wolneykien: Those can't match. :-( More accurately, it depends on the viewer.

Line 88 uses tabs as many other current code, surrounding my changes. (That's why I'd used tabs in my patch in the first place.) Then, I've switched to spaces because you've asked for that — no problem. However, it was necessary to assume a particular tab width to fill in the spaces. I've chosen 8 spaces per tab and the code looks normally indented in my editor. But not here in the GitHub viewer which, probably, uses another tab width.

I think, that's impossible to properly layout the code where tabs and spaces are mixed. Maybe we need to convert all the present tabs to spaces _before_ working on this pull request?

raveit65: You only need to sync line 88 with line 87 and 89.
The problem is that every developer use a different tab size in his editor and MATE is an old code where a lot of people did commit. I have seen files here with 3 different tab sizes. And that's the point where you can't fix it anymore.
For this reason and to avoid this for both sides annoying code-style discussions in nearly 60% of PRs, we decided to use only spaces in newly written code. And in old code it should match the current style, but it has to be look fine at github, because this is our code base.
When nothing helps is is Ok to remove all tabs from a file (in another commit) or from that part where you did edit the file.

wolneykien: > You only need to sync line 88 with line 87 and 89.

…and also line 116, and lines 121--124. So I prefer to cleanup the affected part of the file.

raveit65: Thank you very much.

raveit65 (approved):
LGTM

raveit65:
@rbuj 
Does it work for you?
Any concern left?

helper/pam-helper.c
```diff
@@ -0,0 +1,291 @@
...
+static int
+sane_pam_service(const char *name)
+{
+    const char *sp;
+    char path[128];
+
+    if (strlen(name) > 32)
+        return 0;
+    for (sp = name; *sp; sp++) {
+        if (!isalnum(*sp) && *sp != '_' && *sp != '-')
```
rbuj, line 208: The 1st character of the username should be a-z_, other than first and last may be a-z0-9_-  (optional), and last may be '$' (optional).

helper/pam-helper.c
```diff
@@ -0,0 +1,291 @@
...
+     */
+    while ((fd = open("/dev/null", O_RDWR)) <= 2)
+        ;
+    close(fd);
+
+    /*
+     * Get the program name
+     */
+    if (argc == 0)
+        program_name = "mate-screensaver-pam-helper";
```
rbuj, line 237: This line will never be used, since `argc` should be 3.

helper/pam-helper.c
```diff
@@ -0,0 +1,291 @@
...
+
+    endpwent();
+    return username;
+}
+
+static int
+sane_pam_service(const char *name)
+{
+    const char *sp;
+    char path[128];
```
rbuj, line 203: Use `char path[PATH_MAX];`. See [#include <limits.h>](https://pubs.opengroup.org/onlinepubs/009604499/basedefs/limits.h.html)

helper/pam-helper.c
```diff
@@ -0,0 +1,291 @@
...
+
+    /*
+     * Catch or ignore as many signal as possible.
+     */
+    setup_signals();
+
+    /*
+     * Check argument list
+     */
+    if (argc < 2 || argc > 3) {
```
rbuj, line 251: argc != 3

helper/pam-helper.c
```diff
@@ -0,0 +1,291 @@
...
+     * Discourage users messing around (fat chance)
+     */
+    if (isatty(STDIN_FILENO) && uid != 0) {
+        _log_err(LOG_NOTICE,
+            "Inappropriate use of Unix helper binary [UID=%d]",
+             uid);
+        fprintf(stderr,
+            "This binary is not designed for running in this way\n"
+            "-- the system administrator has been informed\n");
+        sleep(10); /* this should discourage/annoy the user */
```
rbuj, line 275: The sleep time is 10 here.

helper/pam-helper.c
```diff
@@ -0,0 +1,291 @@
...
+    /*
+     * determine the caller's user name
+     */
+    user = getuidname(uid);
+    if (argc == 3 && strcmp(user, argv[2])) {
+        user = argv[2];
+        /* Discourage use of this program as a
+         * password cracker */
+        if (uid != 0)
+            sleep(5);
```
rbuj, line 288: The sleep time is 5 here.

rbuj (commented):
Use a macro for setting the sleep time.

helper/pam-helper.c
```diff
@@ -0,0 +1,291 @@
...
+    char path[128];
+
+    if (strlen(name) > 32)
+        return 0;
+    for (sp = name; *sp; sp++) {
+        if (!isalnum(*sp) && *sp != '_' && *sp != '-')
+            return 0;
+    }
+
+    snprintf(path, sizeof(path), "/etc/pam.d/%s", name);
```
rbuj, line 212: portability issue: hadcoded prefix path

rbuj:
build warnings: `--enable-authentication-scheme=helper`
```
helper_proto.c:37:42: warning: conversion to 'long unsigned int' from 'ssize_t' {aka 'long int'} may change the sign of the result [-Wsign-conversion]
helper_proto.c:83:10: warning: conversion from 'ssize_t' {aka 'long int'} to 'int' may change value [-Wconversion]
helper_proto.c:91:10: warning: conversion from 'ssize_t' {aka 'long int'} to 'int' may change value [-Wconversion]
helper_proto.c:95:15: warning: conversion to 'size_t' {aka 'long unsigned int'} from 'int' may change the sign of the result [-Wsign-conversion]
helper_proto.c:109:43: warning: conversion to 'long unsigned int' from 'ssize_t' {aka 'long int'} may change the sign of the result [-Wsign-conversion]
helper_proto.c:158:12: warning: conversion from 'ssize_t' {aka 'long int'} to 'int' may change value [-Wconversion]
pam-helper.c:78:23: warning: signed conversion from 'unsigned int' to 'int' changes value from '2147483648' to '-2147483648' [-Wsign-conversion]
pam-helper.c:101:41: warning: conversion to 'long unsigned int' from 'int' may change the sign of the result [-Wsign-conversion]
pam-helper.c:123:38: warning: conversion to 'size_t' {aka 'long unsigned int'} from 'ssize_t' {aka 'long int'} may change the sign of the result [-Wsign-conversion]
pam-helper.c:128:43: warning: conversion to 'size_t' {aka 'long unsigned int'} from 'ssize_t' {aka 'long int'} may change the sign of the result [-Wsign-conversion]
pam-helper.c:94:45: warning: unused parameter 'appdata_ptr' [-Wunused-parameter]
gs-auth-helper.c:146:57: warning: passing argument 3 of 'read_prompt' from incompatible pointer type [-Wincompatible-pointer-types]
gs-auth-helper.c:158:42: warning: conversion from 'size_t' {aka 'long unsigned int'} to 'unsigned int' may change value [-Wconversion]
gs-auth-helper.c:164:36: warning: format '%d' expects argument of type 'int', but argument 4 has type 'ssize_t' {aka 'long int'} [-Wformat=]
gs-auth-helper.c:187:40: warning: unused parameter 'display' [-Wunused-parameter]
gs-auth-helper.c:190:40: warning: unused parameter 'error' [-Wunused-parameter]
```
```
#!/bin/bash
export LANG=en_US.UFT-8

NPROC=`nproc`
CFLAGS="-march=native -mtune=native -fomit-frame-pointer -malign-double -fstrict-aliasing -ffast-math -Wconversion -Wunused-macros -Wunused-parameter -Wdiscarded-qualifiers"
CONF="--enable-authentication-scheme=helper --prefix=/usr --enable-compile-warnings=maximum --enable-cxx-warnings=maximum --disable-debug"
CFLAGS="${CFLAGS}" ./autogen.sh ${CONF} &> conf.log && make -j${NPROC} &> make.log && sudo make install &> /dev/null
```

rbuj (commented):
I get a blank screen after locking the screen using the builtin helper on Fedora
The user/password dialog isn't displayed.
I need to restart the session to come back to X:
```
$ sudo systemctl restart lightdm.service
```

helper/pam-helper.c
wolneykien: This is about PAM service name (login, ssh, lightdm, ...), not a user name.

wolneykien:
> I get a blank screen after locking the screen using the builtin helper on Fedora
> The user/password dialog isn't displayed.

I'll check it on Debian.

wolneykien:
> > I get a blank screen after locking the screen using the builtin helper on Fedora
> > The user/password dialog isn't displayed.
> 
> I'll check it on Debian.

1. Fresh Debian 10 installation with MATE desktop.
2. Switch to "bullseye" repo (Debian testing), `apt-get update`, `apt-get dist-upgrade`, `reboot`. It might be ok to just install the Debian testing `.iso`.
3. `apt-get install intltool libdbus-glib-1-dev libdconf-dev libgl1-mesa-dev libglib2.0-dev libgtk-3-dev libmate-desktop-dev libmate-menu-dev libmatekbd-dev libnotify-dev libpam0g-dev librda-dev libstartup-notification0-dev libsystemd-dev libx11-dev libxext-dev libxss-dev libxt-dev libxxf86vm-dev mate-common x11proto-scrnsaver-dev x11proto-xext-dev x11proto-xf86vidmode-dev xmlto autopoint make`
4. `apt-get install git-core`
5. `git clone https://github.com/wolneykien/mate-screensaver.git`
6. `cd mate-screensaver/`
7. `git checkout announce`
8. `NOCONFIGURE=1 ./autogen.sh`
9. `./configure --prefix=/usr --disable-silent-rules --disable-static --with-xscreensaverdir=/usr/share/xscreensaver/config --with-xscreensaverhackdir=/usr/lib/xscreensaver --with-mit-ext --libexecdir=/usr/lib/mate-screensaver --with-libnotify --enable-locking --enable-pam --enable-authentication-scheme=helper --with-systemd --without-console-kit`
10. `make`
11. `make install DESTDIR=$HOME/tmp`
12. Don't override the whole screensaver installation (as it may be incompatible with the Debian packaging scheme), just copy the two binaries: `usr/lib/mate-screensaver/mate-screensaver` and `usr/lib/mate-screensaver/mate-screensaver/mate-screensaver-pam-helper` from `$HOME/tmp/` to the system.
13. `chmod u+s /usr/lib/mate-screensaver/mate-screensaver/mate-screensaver-pam-helper` (it should be **suid root**).
14. Lock the screen. Press some key to activate the dialog.
15. Found `/usr/lib/mate-screensaver/mate-screensaver-dialog` and `/usr/lib/mate-screensaver/mate-screensaver/mate-screensaver-pam-helper` in the process tree.
16. Unlock the screen by entering the password.
17. Restart the MATE session and repeat steps 14--16.

It seems to work.

wolneykien:
**About testing on Fedora:** I think the critical step is number **13** above: suid root the helper binary. Then, if `/etc/pam.d/mate-screensaver` is present, it should function properly.

rbuj:
> **About testing on Fedora:** I think the critical step is number **13** above: suid root the helper binary. Then, if `/etc/pam.d/mate-screensaver` is present, it should function properly.

The variable `${exec_prefix}` wasn't replaced on the configure steep:
```
/usr/libexec/mate-screensaver-dialog

** (process:4879): WARNING **: 14:15:06.008: ${exec_prefix}/libexec/mate-screensaver/mate-screensaver-pam-helper does not exist. password authentication via external helper will not work.
NOTICE=AUTH FAILED
NOTICE=AUTH FAILED
NOTICE=AUTH FAILED
NOTICE=AUTH FAILED
NOTICE=AUTH FAILED
REQUEST QUIT
```

Test:
```
$ grep PASSWD_HELPER_PROGRAM config.h
#define PASSWD_HELPER_PROGRAM "${exec_prefix}/libexec/mate-screensaver/mate-screensaver-pam-helper"
```

wolneykien:
> The variable `${exec_prefix}` wasn't replaced on the configure steep:
> ...
> ```
> #define PASSWD_HELPER_PROGRAM "${exec_prefix}/libexec/mate-screensaver/mate-screensaver-pam-helper"
> ```

It seems like the following:
* https://www.howtobuildsoftware.com/index.php/how-do/dy5/autotools-autoconf-automake-m4-ac-subst-does-not-expand-variable
* https://stackoverflow.com/questions/24137728/define-a-variable-in-configure-ac-as-output-of-a-command

I'll try to workaround that, but it depends on how `libexecdir` is defined by default. Personally, I don't have this problem due to explicit `--with-libexecdir=...` passed to `./configure`.

wolneykien:
I've redefined `PASSWD_HELPER_PROGRAM` the following way:

```diff
-    AC_DEFINE_UNQUOTED(PASSWD_HELPER_PROGRAM, "${libexecdir}/mate-screensaver/mate-screensaver-pam-helper", [Full pathname of password helper application])
+    AC_DEFINE_UNQUOTED(PASSWD_HELPER_PROGRAM, [PKGLIBEXECDIR "/mate-screensaver-pam-helper"], [Full pathname of password helper application])
```

`-DPKGLIBEXECDIR=...` is defined at _build time_ and seems to be correct both by default and with `.\configure --libexecdir=...`.

wolneykien:
`strings src/mate-screensaver-dialog | grep 'helper'` shows that the path is fully expanded --- no variables.

rbuj:
> I've redefined `PASSWD_HELPER_PROGRAM` the following way:
> 
> ```diff
> -    AC_DEFINE_UNQUOTED(PASSWD_HELPER_PROGRAM, "${libexecdir}/mate-screensaver/mate-screensaver-pam-helper", [Full pathname of password helper application])
> +    AC_DEFINE_UNQUOTED(PASSWD_HELPER_PROGRAM, [PKGLIBEXECDIR "/mate-screensaver-pam-helper"], [Full pathname of password helper application])

It makes sense using the directory variable `PKGLIBEXECDIR` after setting `pkglibexec_PROGRAMS`:
```
$ git grep pkglibexec_PROGRAMS
helper/Makefile.am:pkglibexec_PROGRAMS = mate-screensaver-pam-helper
```

wolneykien:
> It makes sense using the directory variable `PKGLIBEXECDIR` after setting `pkglibexec_PROGRAMS`:
> 
> ```
> $ git grep pkglibexec_PROGRAMS
> helper/Makefile.am:pkglibexec_PROGRAMS = mate-screensaver-pam-helper
> ```

Hm... What do you mean?

Currently, `pkglibexecdir` is defined in the two `Makefile.am` files _independently_:
1. in `src/Makefile.am` where `-DPKGLIBEXECDIR=$(pkglibexecdir)` is added to the `CFLAGS` (making it possible to expand the `PASSWD_HELPER_PROGRAM` macro and to compile `src/gs-auth-helper.c`) and
2. in `helper/Makefile.am` where it is defined to make `pkglibexec_PROGRAMS` work properly.

The obvious way to optimize that is to define `pkglibexecdir` only once, in the top-level `Makefile.am`. However, I don't see why, how and where to additionally use `PKGLIBEXECDIR` --- the C macro I've introduced to workaround the variable expansion issue with `$exec_prefix`.

rbuj:
> The obvious way to optimize that is to define `pkglibexecdir` only once, in the top-level `Makefile.am`.

You can define the variable [pkglibexecdir](https://www.gnu.org/software/automake/manual/html_node/Uniform.html#index-pkglibexecdir) in the configure.ac file.

wolneykien:
> You can define the variable [pkglibexecdir](https://www.gnu.org/software/automake/manual/html_node/Uniform.html#index-pkglibexecdir) in the configure.ac file.

Actually, it is already defined there as `privlibexecdir`. I'll try to use it.

rbuj:
> > You can define the variable [pkglibexecdir](https://www.gnu.org/software/automake/manual/html_node/Uniform.html#index-pkglibexecdir) in the configure.ac file.
> 
> Actually, it is already defined there as `privlibexecdir`. I'll try to use it.

`pkglibexecdir` is already available without explicitly defining it.
```
$ grep -R -F "pkglibexecdir =" --include=Makefile.in
data/images/cosmos/Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
data/images/Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
data/Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
doc/Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
savers/Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
src/Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
helper/Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
Makefile.in:pkglibexecdir = $(libexecdir)/@PACKAGE@
```

wolneykien:
> `pkglibexecdir` is already available without explicitly defining it.

Done. `pkglibexecdir` is used in `src/Makefile.am` and `helper/Makefile.am` without an explicit definition.

wolneykien:
Hi. Any pending questions?

raveit65:
I will test it again, but we have to wait when @rbuj  is back.

raveit65:
For some reason i can't pull anymore the changes from you branch.
```
[rave@mother mate-screensaver]$ git checkout -b wolneykien-announce master
Switched to a new branch 'wolneykien-announce'
[rave@mother mate-screensaver]$ git pull https://github.com/wolneykien/mate-screensaver.git announce
From https://github.com/wolneykien/mate-screensaver
 * branch            announce   -> FETCH_HEAD
fatal: Not possible to fast-forward, aborting.
[rave@mother mate-screensaver]$
```

raveit65:
Ok, after forking your repo to my account an cloning it to my local box i was able to get your changes again.

raveit65:
@rbuj 
Any concern left or do you like to approve it?

wolneykien:
I think it's ok. However, I forgot to test it for _backward_ compatibility with the previous helper. I'll make that ASAP.

src/gs-auth-helper.c
```diff
@@ -50,6 +50,9 @@
 #include "gs-auth.h"
 #include "subprocs.h"
 
+#include "../helper/helper_proto.h"
```
rbuj, line 53: Do not use relative paths for headers, add a search path to the file using "-I"

Makefile.am
```diff
@@ -9,7 +9,12 @@ DISTCHECK_CONFIGURE_FLAGS = \
 	--without-systemd \
 	CFLAGS='-Wno-deprecated-declarations'
 
-SUBDIRS =			\
+SUBDIRS =
+if BUILTIN_HELPER
+SUBDIRS += helper
+endif
+
+SUBDIRS +=	\
```
rbuj, line 17: There is any reason why the helper folder cannot be attached after setting fixed folders of SUBDIRS.

helper/pam-helper.c
rbuj: ping

helper/pam-helper.c
rbuj: ping

helper/pam-helper.c
rbuj: ping

helper/pam-helper.c
rbuj: ping

helper/pam-helper.c
rbuj: macro for setting an unique delay value?

rbuj:
Did you fix all new warnings listed at https://github.com/mate-desktop/mate-screensaver/pull/178#issuecomment-797228438?

rbuj:
[tbd](https://dictionary.cambridge.org/es/diccionario/ingles/tbd):
- https://github.com/mate-desktop/mate-screensaver/pull/178#discussion_r592891975
- https://github.com/mate-desktop/mate-screensaver/pull/178#discussion_r592894014
- https://github.com/mate-desktop/mate-screensaver/pull/178#discussion_r592894578
- https://github.com/mate-desktop/mate-screensaver/pull/178#discussion_r592895176 / https://github.com/mate-desktop/mate-screensaver/pull/178#discussion_r592895444
- https://github.com/mate-desktop/mate-screensaver/pull/178#discussion_r592900315

rbuj (approved):
I don't want to block it... it would be great if another else could review it.

wolneykien:
> I don't want to block it... it would be great if another else could review it.

I'll react to your comments above soon.

rbuj:
> > I don't want to block it... it would be great if another else could review it.
> 
> I'll react to your comments above soon.

ok, don't worry

lukefromdc:
What is the status of this?

helper/pam-helper.c
wolneykien: The third argument is currently optional, see line 282.

helper/pam-helper.c
wolneykien: Why do we need an unique (random?) delay here?

helper/pam-helper.c
wolneykien: The whole `sane_pam_service()` function doesn't seem to be portable. Thanks.

helper/pam-helper.c
wolneykien: https://github.com/wolneykien/mate-screensaver/commit/66a3493d8868905602140158d75b10c904354b9f

helper/pam-helper.c
wolneykien: https://github.com/wolneykien/mate-screensaver/commit/3dcbd1ef8482f1847245627fc01b42888cf0b080

helper/pam-helper.c
wolneykien: https://github.com/wolneykien/mate-screensaver/commit/c681ca8ebfee1b346e17eed5502f70ceba8f404d

raveit65:
@rbuj any chance to finish this review?

raveit65:
I will start testing it again.

helper/helper_proto.c
```diff
@@ -0,0 +1,159 @@
+/* Part of xscreensaver.
+ * Copyright (c) 2019 Paul Wolneykien <manowar@altlinux.org>
```
vkareh, line 2: Why is this marked as "Part of xscreensaver" - I don't see any of this code in the latest source for xscreensaver. Also, if this is the case, is the Copyright attribution correct?

lukefromdc:
As I recall, gnome-screensaver's code began life with a fork of xscreensaver so this might just be original code.

On 7/9/2021 at 9:21 AM, "Victor Kareh" ***@***.***> wrote:
>
***@***.*** commented on this pull request.

vkareh:
@lukefromdc my question is about attribution: jwz (Jamie Zawinski) owns the copyright for xscreensaver, not @wolneykien (Paul Wolneykien). So if this file is in fact part of xscreensaver, then the attribution needs to be fixed (and we're in possible license violation, as xscreensaver is BSD). If the file is _not_ part of xscreensaver, then that message needs to be removed.

lukefromdc:
I don't know the answer to that offhand,save that the license issue led to a discussion that went nowhere on the grounds that it applies equally to gnome-screensaver and cinnamon-screensaver.There was discussion of contacting the xscreensaver devs about relicensing but nothong after that-and no response known to me from xscreensaver devs on forums where this came up

On 7/12/2021 at 8:05 AM, "Victor Kareh" ***@***.***> wrote:
> 
***@***.*** my question is about attribution: jwz (Jamie Zawinski)

helper/helper_proto.c
wolneykien: Sorry, @vkareh , I too don't know why marked it that way. :-) This PR is quite old! 
However, that mistake can have relations with the copyright message of `src/gs-auth-helper.c`:

```
 * written by Olaf Kirch <okir@suse.de>
 * xscreensaver, Copyright (c) 1993-2004 Jamie Zawinski <jwz@jwz.org>
```

No more idea. I'll replace it with "a part of mate-screensaver", ok?

vkareh: If it's your original code, of course! You can mark it as part of mate-screensaver or whatever, with copyright to your name :)

Just make sure that it doesn't reference xscreensaver (or its code) without attribution to jwz.

wolneykien: I'm sure the two-way protocol is my original code. By examining [`driver/passwd-helper.c`](https://github.com/Zygo/xscreensaver/blob/master/driver/passwd-helper.c) from XScreenSaver we can see that it resembles `gs-auth-helper.c` of the current `mate-screensaver` (and that it isn't actually developed any more).

See my next comments with questions about licensing.

wolneykien:
**Many thanks to @vkareh for rising the question about the licensing of new code in this PR!** (For an unknown reason the `helper/helper_proto.*` is attributed now to XScreenSaver (no relation actually). And I'll fix that soon.)

The question now is the following: what license should be selected for `helper/helper_proto.*`? I see that `mate-screensaver` is normally licensed under GNU GPL. However, the `gs-auth-helper.c`, which uses `helper/helper_proto.*`, is licensed under a BSD-like license.

raveit65:
> **Many thanks to @vkareh for rising the question about the licensing of new code in this PR!** (For an unknown reason the `helper/helper_proto.*` is attributed now to XScreenSaver (no relation actually). And I'll fix that soon.)
> 
Please do.

> The question now is the following: what license should be selected for `helper/helper_proto.*`? I see that `mate-screensaver` is normally licensed under GNU GPL. However, the `gs-auth-helper.c`, which uses `helper/helper_proto.*`, is licensed under a BSD-like license.

Please use GNU GPL like the other files of mate-screensaver.
No need to make ist complicate at this point.
Can you please rebase PR to fix conflict with master?
Code changes are looking good and are tested for a long time by me.
So i will merge it.

wolneykien:
> Code changes are looking good and are tested for a long time by me.
> So i will merge it.

Thanks! I've updated the copyright messages. If you want, I can squash the commits into one.

raveit65:
Yes please, this makes it easier for a final build & run test.

helper/pam-helper.c
```diff
@@ -0,0 +1,287 @@
...
+static int
+_converse(int num_msg, const struct pam_message **msg,
+          struct pam_response **resp, void *appdata_ptr)
+{
+    struct pam_response *reply;
+    char buf[MAXLEN];
+    int num;
+    int ret = PAM_SUCCESS;
+
+    if (!(reply = malloc(sizeof(*reply) * num_msg)))
```
mbkma, line 120: where do you free the memory allocated by this malloc call?

helper/pam-helper.c
mbkma: why not change this to `argc != 3`? It is equivalent.

helper/pam-helper.c
raveit65: @wolneykien 
Here is another open Question by a team member.

wolneykien: According to https://linux.die.net/man/3/pam_conv , the `pam_response` is normally freed by the PAM module (the caller in that case). However, in the case of an error, _"the conversation function should release any resources it has allocated"_, so you're rigjt, @mbkma, I need to add `free(reply)` to the error case branch there below.

wolneykien: Hi! I've just fixed that.

wolneykien:
> Yes please, this makes it easier for a final build & run test.

If the last changes are OK, then I'll squash the commits.

raveit65:
Good, let's merge it. I will cherry-pick it to 1.26 branch. No need to wait another year to see this code in production.

raveit65:
Ok, my fault, pam-devel wasn't installed.

PR #179, opened 2019-01-25
1.20

dagobertstyle:
Hi, here is the 1.20 version...but still no way to change the personnal images folder with the GUI ?


raveit65:
What the hell is that?
Please don't do any PR for stable branch.

PR #181, opened 2019-02-13
Make translatable the auth-username-label in lock-dialog-default.ui

rbuj:


sc0w:
I think this is the affected word:



It must to be translatable

PR #188, opened 2019-03-27
Add ELOGIND support

### LINKED ISSUE #187: Systemd question
In gs-listner-dbus.c there is:
`

					/* Instead of going via the
					 * bus to read the new
					 * property state, let's
					 * shortcut this and ask
					 * directly the low-level
					 * information */

					new_active = sd_session_is_active (listener->priv->session_id) != 0;`

This appear to be the only build dep for systemd and I wonder why you need it?

vkareh:
Reuses WITH_SYSTEMD #define to reduce #ifdef mess.

ELOGIND is not officially supported, but since it's
a subset of systemd, it should work out of the box.

Co-authored-by: Joakim Tjernlund <joakim.tjernlund@infinera.com>

Fixes #187

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -945,6 +945,32 @@ fi
 AC_SUBST(SYSTEMD_CFLAGS)
 AC_SUBST(SYSTEMD_LIBS)
 
+dnl ====================================================================
+dnl Check for elogind
+dnl
+dnl elogind is not officially supported, but since it's a subset of
+dnl systemd, it should work out of the box
+dnl ====================================================================
+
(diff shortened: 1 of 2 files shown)


vkareh:
@joakim-tjernlund - Since elogind is not officially supported, I've updated your patch to disable the `--with-elogind=auto` mode and default to `no`. I also removed an unnecessary duplicate check for libelogind. Thanks for this patch!

joakim-tjernlund:
@vkareh - Good, I am happy with these changes, Thanks

joakim-tjernlund:
ping?

Issue #175, opened 2019-01-15
Black screen when suspending or hibernating via gui


bwat47:
#### Expected behaviour
Upon resume from suspend or hibernate, I should see the lock screen without having to touch my mouse or keyboard

#### Actual behaviour
Upon resume from suspend or hibernate I have a black screen (the screen is clearly on, but it's displaying all back, as if the 'blank' screensaver is being displayed). The lock screen/password dialogue will not appear until I either move my mouse or hit a key on the keyboard.

NOTE: I ONLY see this issue occur if I suspend (or hibernate) via the mate 'shut down the computer' dialogue. If I instead suspend the system by closing my laptop lid, it works as expected upon resume (the lock screen appears right away)

#### Steps to reproduce the behaviour
1. Add the MATE 'Shut Down' applet to the panel (if not already)
2. Click the Shut Down the Computer icon on the panel
3. From the 'Shutdown this system now' dialogue, select Suspend
4. After the machine suspends, power it back in

#### MATE general version
1.20.4

#### Package version
mate-screensaver 1.20.3

#### Linux Distribution
Solus

bwat47:
Also, if it helps I attached the relevant mate lock screen settings on my machine:

![screenshot at 2019-01-14 20-59-47](https://user-images.githubusercontent.com/1442520/51153798-8121e800-183f-11e9-81bc-f7ffa2a06c54.png)

raveit65:
I couldn't reproduce this issue with mate-screensaver-1.22.1 running on my MSI FX700 notebook with fedora 30.
Here i see always the default wallpaper after suspend-to-disk or suspend-to ram, and the lock-dialog appears after 2-3 secs.
Any way, might be different on different hardware.

monsta:
This might be fixed now by https://github.com/mate-desktop/mate-screensaver/issues/175, but I can't reproduce this, so I'm not sure.

DHLF:
While running some steam games using proton it seems if I stay away from keyboard too long (roughly some 10 minutes or so) the games freezes in such a bad way.. not even desktop switching was possible.

The sceen was blanked(black) until i touch a key or move the mouse around. 

The mousepointer was there(moveable) and as background the image of the (previous running) game could be seen. No button could be clicked. No response.

Using console ctrl+f1 i tried the following stuff:
sudo ps aux | grep creens   (shortened because i dont knew if it will be Screensaver or screensaver..
[...]

Issue #145, opened 2017-12-15
Lock screen appears late after wakeup from suspend


omgold:
#### Expected behaviour

When the power manager / screensaver is configured to lock the desktop when suspending the system, the display should come up from system with the lock screen or screensaver already covering the desktop.

#### Actual behaviour

After wakeup from suspend it takes a few seconds, with the desktop shown, before the lock screen appears.

Note: it doesn't happen when screensaver them is set to blank

#### Steps to reproduce the behaviour

- set lock-suspend in power manager config to true
- set lock-enabled in screensaver config to true
- set screensaver theme to something other than blank screen
- suspend system (while normal desktop is shown, not screensaver or lock screen)
- wake up system (best with monitor not having gone to standby meanwhile)

#### MATE general version

1.18.0

#### Package version

mate-power-manager 1.18.1
mate-screensaver 1.18.2

#### Linux Distribution

Arch

grisu48:
I can confirm the issue, I'm having the same problem with the same packages
* community/mate-power-manager 1.18.1-1 (mate-extra)
* community/mate-screensaver 1.18.2-1 (mate-extra)

MingcongBai:
Another confirmation, AOSC OS: MATE Screensaver 1.18.2.

bwat47:
I can confirm this here.

IMO it would make more sense if the lock screen activated *before* suspend (and suspend was inhibited until the lock screen is up), I believe this is what gnome-shell and cinnamon does

cu:
I see the same issue on Ubuntu MATE 18.04.

lukefromdc:
I saw an interesting discussion on these sort of issues concerning of all things Wayland. Apparently, on X screensavers have to operate by drawing over all other windows and intercepting all keypresses etc. Thus, if anything causes that window to appear late, close when it should not, not grab the keyboard, etc, the result is a security hole in the screensaver. The hope when that story was written was that Wayland would fix that, no idea if that ever panned out or not. Anyway, this may be one reason these kinds of issues keep cropping up with screensavers. 

Given that, I recommend at least logging out on any Linux system (really any system) if it must be left unattended in an untrusted environment. My encrypted systems get shut all the way down in that case due to the inability of Linux encryption to protect against the "cold boot attack" to recover keys from RAM.

bwat47:
I've used some X screenlockers that don't seem to have this issue (or at least have it far less frequently). Namely, gnome-shell's lock screen and light-locker.

[...]

PR #202, opened 2019-07-31
Implement lock on suspend

apaan:
This patch fixes two issues #145 and #175. 

It is adopted from xfce4-screensaver patch:
https://git.xfce.org/apps/xfce4-screensaver/commit/?id=9e53bb2866

Introduced new Inhibitor lock to make sure lock was in place and
user will be greeted with unlock dialog when resumes from suspend/
hibernate.

### DIFF
--- a/src/gs-listener-dbus.c
+++ b/src/gs-listener-dbus.c
@@ -96,7 +96,8 @@ struct GSListenerPrivate
 	char           *session_id;
 
 #ifdef WITH_SYSTEMD
-	gboolean have_systemd;
+	gboolean        have_systemd;
+	gint            logind_inhibit_lock;
 #endif
 
 	guint32         ck_throttle_cookie;
@@ -122,7 +123,8 @@ enum
     ACTIVE_CHANGED,
     THROTTLE_CHANGED,
     SHOW_MESSAGE,
-    LAST_SIGNAL
+    PREPARE_FOR_SLEEP,
+    LAST_SIGNAL,
 };
 
 enum
@@ -1619,6 +1621,66 @@ failure:
 	gs_debug ("Failed to decode PropertiesChanged message.");
 	return FALSE;
 }
+
+static gint
+take_logind_inhibit_lock (DBusConnection *connection)
+{
+	DBusMessage    *message;
+	DBusMessage    *reply;
+	DBusError       error;
+	DBusMessageIter reply_iter;
+	gint            fd;
+
+	const char* what = "sleep";
+	const char* who  = g_get_user_name ();
+	const char* why  = "Lock screen before sleep";
+	const char* mode = "delay";
+
+	g_return_val_if_fail (connection != NULL, FALSE);
+
+	dbus_error_init (&error);
+
+	message = dbus_message_new_method_call (SYSTEMD_LOGIND_SERVICE,
+	                                        SYSTEMD_LOGIND_PATH,
+	                                        SYSTEMD_LOGIND_INTERFACE,
+	                                        "Inhibit");
+	if (message == NULL)
+	{
+		gs_debug ("Couldn't allocate the dbus message");
+		return 0;
+	}
+
+	if (dbus_message_append_args (message,
+	                              DBUS_TYPE_STRING, &what,
+	                              DBUS_TYPE_STRING, &who,
+	                              DBUS_TYPE_STRING, &why,
+	                              DBUS_TYPE_STRING, &mode,
+	                              DBUS_TYPE_INVALID) == FALSE)
+	{
+		gs_debug ("Couldn't add args to the dbus message");
+		return 0;
+	}
+
+	reply = dbus_connection_send_with_reply_and_block (connection, message,
+	                                                  -1, &error);
+	dbus_message_unref (message);
+
+	if (dbus_error_is_set (&error))
+	{
+		gs_debug ("%s raised:\n %s\n\n", error.name, error.message);
+		dbus_error_free (&error);
+		return 0;
+	}
+
+	dbus_message_iter_init (reply, &reply_iter);
+	dbus_message_iter_get_basic (&reply_iter, &fd);
+
+	dbus_message_unref (reply);
+
+	gs_debug ("System inhibitor fd is %d\n", fd);
+
+	return fd;
+}
 #endif
 
 static DBusHandlerResult
@@ -1655,6 +1717,28 @@ listener_dbus_handle_system_message (DBusConnection *connection,
 				gs_debug ("systemd requested session lock");
 				g_signal_emit (listener, signals [LOCK], 0);
 			}
+
+			return DBUS_HANDLER_RESULT_HANDLED;
+		} else if (dbus_message_is_signal (message, SYSTEMD_LOGIND_INTERFACE, "PrepareForSleep")) {
+			gboolean  active = 0;
+			DBusError error;
+
+			dbus_error_init (&error);
+			dbus_message_get_args (message, &error, DBUS_TYPE_BOOLEAN, &active, DBUS_TYPE_INVALID);
+			if (active) {
+				gs_debug ("Logind wanted to sleep");
+				g_signal_emit (listener, signals [PREPARE_FOR_SLEEP], 0, active);
+				if (listener->priv->logind_inhibit_lock) {
+					gs_debug ("Releasing inihibitor lock");
+					close (listener->priv->logind_inhibit_lock);
+					listener->priv->logind_inhibit_lock = 0;
+				}
+			} else {
+				gs_debug ("Logind resumed from sleep");
+				listener->priv->logind_inhibit_lock = take_logind_inhibit_lock (listener->priv->system_connection);
+				g_signal_emit (listener, signals [PREPARE_FOR_SLEEP], 0, active);
+			}
+			gs_debug ("Logind PrepareForSleepHandled");
 
 			return DBUS_HANDLER_RESULT_HANDLED;
 		} else if (dbus_message_is_signal (message, DBUS_INTERFACE_PROPERTIES, "PropertiesChanged")) {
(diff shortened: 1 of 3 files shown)


lukefromdc:
Can users turn this on and off with a gsetting? Not everyone wants the screen locked when coming out of suspend. Some need is and some do not.

The obvious example of this wouldbe many users of encrypted laptops, who must maintain absolute custody of the machine anyway whenever it is booted as cold boot attacks and other means can recover keys from encrypted computers stolen/confiscated while running and decrypted.

apaan:
Testing on my laptop, without this patch, mate-screensaver behaves according to option "Lock screen when screensaver is active" in Screensaver Preferences (stored in ```lock-enabled``` gsetting):
- when the option is active, the screen will lock on suspend/hibernate (screensaver is activated too)
- when the option is inactive, the screen will not lock on suspend (screensaver is not activated, even if the suspend time is longer than "Idle time" option)

If this is correct, i will modify the patch so that the behavior will not change.

Thanks

lukefromdc:
The described behavior in master "Lock screen when screensaver is active" should be the correct behavior and is simpler than adding a new and separate action.

src/gs-listener-dbus.c
```diff
@@ -1656,6 +1751,27 @@ listener_dbus_handle_system_message (DBusConnection *connection,
...
+			if (listener->priv->lock_enabled) {
+				gboolean  active = 0;
+				DBusError error;
+
+				dbus_error_init (&error);
+				dbus_message_get_args (message, &error, DBUS_TYPE_BOOLEAN, &active, DBUS_TYPE_INVALID);
+				if (active) {
+					gs_debug ("Logind wanted to sleep");
+					g_signal_emit (listener, signals[LOCK], 0);
+					g_signal_emit (listener, signals[SHOW_MESSAGE], 0, NULL, NULL, NULL);
```
monsta: What does `SHOW_MESSAGE` signal do in this case...?

monsta:
Xfce's screensaver is a fork of this one... how does it work without that new inhibitor?

monsta:
> The described behavior in master "Lock screen when screensaver is active" should be the correct behavior and is simpler than adding a new and separate action.

I recall we have a bunch of "lock-on-something" keys in mate-power-manager schema. However, it's probably only related to cases when suspend is activated on (hardware) button press or on closing the laptop lid...

src/gs-listener-dbus.c
apaan: To bring up unlock dialog, because we don't have "show unlock" signal.
This is done twice (one before suspend, and one after suspend) because there is a probability that the suspend process is taking long time, and the dialog is closed before suspend. (see issue #175 )

apaan:
> Xfce's screensaver is a fork of this one... how does it work without that new inhibitor?

Actually I started by copying the Xfce's patch (I even put a link to it in the commit message). But it didn't always work on my notebook. So i dug logind documentation and found this (taken from https://www.freedesktop.org/wiki/Software/systemd/inhibit/):

> Note that watching PrepareForShutdown(true)?/PrepareForSleep(true) without taking a delay lock is racy and should not be done, as any code that an application might want to execute on this signal might not actually finish before the suspend/shutdown cycle is executed. Again: if you watch PrepareForSuspend(true), then you really should have taken a delay lock first.

apaan:
> > The described behavior in master "Lock screen when screensaver is active" should be the correct behavior and is simpler than adding a new and separate action.
> 
> I recall we have a bunch of "lock-on-something" keys in mate-power-manager schema. However, it's probably only related to cases when suspend is activated on (hardware) button press or on closing the laptop lid...

Just checked that. mate-power-manager has ```lock-use-screensaver```: Whether to use the screen lock setting of mate-screensaver to decide if the screen is locked after a hibernate, suspend or blank screen.

Current behavior is different from master in this regard. Will look into it later

monsta:
Ok, so we delay the suspend until we display the lock dialog, and then release it...

> Note that watching PrepareForShutdown(true)?/PrepareForSleep(true) without taking a delay lock is racy and should not be done, as any code that an application might want to execute on this signal might not actually finish before the suspend/shutdown cycle is executed. Again: if you watch PrepareForSuspend(true), then you really should have taken a delay lock first.

So that's why we also acquire the lock after resuming? To play it safe for the next PrepareForSleep...

monsta:
Interesting that Xfce's screensaver still doesn't have neither inhibit nor showing the dialog before the suspend...

raveit65:
> So that's why we also acquire the lock after resuming? To play it safe for the next PrepareForSleep...

Why not simply testing the PR, it's self explanatory....

apaan:
> > The described behavior in master "Lock screen when screensaver is active" should be the correct behavior and is simpler than adding a new and separate action.
> 
> I recall we have a bunch of "lock-on-something" keys in mate-power-manager schema. However, it's probably only related to cases when suspend is activated on (hardware) button press or on closing the laptop lid...

Pushed a change: This patch now lets mate-power-manager to lock or not. This patch only ensures that the first thing that we see on resume is unlock dialog. Same behavior as before, but now, this patch behaves correctly on hidden mate-power-manager settings.

src/gs-listener-dbus.c
```diff
@@ -1619,6 +1621,66 @@ properties_changed_match (DBusMessage *message,
...
+	const char* who  = g_get_user_name ();
+	const char* why  = "Lock screen before sleep";
+	const char* mode = "delay";
+
+	g_return_val_if_fail (connection != NULL, FALSE);
+
+	dbus_error_init (&error);
+
+	message = dbus_message_new_method_call (SYSTEMD_LOGIND_SERVICE,
+											SYSTEMD_LOGIND_PATH,
```
monsta: Minor thing: indentation is way off...

src/gs-listener-dbus.c
```diff
@@ -1656,6 +1718,27 @@ listener_dbus_handle_system_message (DBusConnection *connection,
...
+			DBusError error;
+
+			dbus_error_init (&error);
+			dbus_message_get_args (message, &error, DBUS_TYPE_BOOLEAN, &active, DBUS_TYPE_INVALID);
+			if (active) {
+				gs_debug ("Logind wanted to sleep");
+				g_signal_emit (listener, signals [PREPARE_FOR_SLEEP], 0, active);
+				if (listener->priv->logind_inhibit_lock) {
+					gs_debug ("Releasing inihibitor lock");
+					close (listener->priv->logind_inhibit_lock);
```
monsta: Hmm... `close` won't set the descriptor to zero.

src/gs-monitor.c
```diff
@@ -280,6 +280,17 @@ static void listener_simulate_user_activity_cb(GSListener* listener, GSMonitor*
 	gs_monitor_simulate_user_activity(monitor);
 }
 
+static void listener_prepare_for_sleep_cb(GSListener* listener, gboolean prepare, GSMonitor* monitor)
+{
+	gboolean locked;
```
monsta: Should we also check if `prepare` is `TRUE` here?

raveit65:
@apaan 
I found the working version of your work.
https://github.com/mate-desktop/mate-screensaver/compare/410fcb6fcbff537ba0cb24b834a86c68721cdb7d..4dbbbe7b8e72e3a54b5d0657a94b794bcee1ab04.patch
Thanks for that :)
I will use that for fedora and step back from the review here, because I have no idea why reviewers breaks your work here with their ideas....

src/gs-listener-dbus.c
```diff
@@ -1619,6 +1621,66 @@ properties_changed_match (DBusMessage *message,
...
+		gs_debug ("Couldn't allocate the dbus message");
+		return 0;
+	}
+
+	if (dbus_message_append_args (message,
+						DBUS_TYPE_STRING, &what,
+						DBUS_TYPE_STRING, &who,
+						DBUS_TYPE_STRING, &why,
+						DBUS_TYPE_STRING, &mode,
+						DBUS_TYPE_INVALID) == FALSE)
```
vkareh: Indentation here is off as well

src/gs-listener-dbus.c
```diff
@@ -1619,6 +1621,66 @@ properties_changed_match (DBusMessage *message,
...
+	{
+		gs_debug ("Couldn't add args to the dbus message");
+		return 0;
+	}
+
+	reply = dbus_connection_send_with_reply_and_block (connection, message,
+		-1, &error);
```
vkareh: Indentation...

vkareh:
Still need to figure out why it doesn't work for @raveit65 - that's a potential blocker...

monsta:
This also works for me if I disable locking in screensaver settings, but explicitly tell the power manager to ask the screensaver to lock on suspend.

```
$ gsettings get org.mate.screensaver lock-enabled
false
$ gsettings get org.mate.power-manager lock-use-screensaver
false
$ gsettings get org.mate.power-manager lock-suspend
true
```
Suspend was tested by setting the power button to perform suspend in power manager settings, then pressing the power button (obviously).

monsta:
The commit message should probably be changed, because lock-on-suspend is already implemented, and this change just ensures it...

raveit65:
Ok, i was able to reproduce the issue and i don't think it is related to PR.
On a system with 2 monitors (non laptop) is is possible that the lock-dialog is on a monitor which isn't powered on after coming from suspend-to-ram = ***Computer are stupid!***
Than you see only a black screen at the monitor which is on.
In details:
1. Both monitors are enabled in mate-window-properties
2. Suspend the system
3. Plug off the second monitor from electrical net to save power.
4. Go shopping  (forget everything!)
5. Come back from shopping, and start system from suspend with one monitor.
6. Have a black screen without a lock-dialog.

Note, those steps aren't always reproducible and i think this is a monitor detection issue with a multi-monitor setup, not related to mate-screensaver.

src/gs-listener-dbus.c
apaan: ouch. Thanks. Failed to see that the file is indented by tabs. Will correct it in the next patch (with spaces)

src/gs-listener-dbus.c
apaan: Thanks

src/gs-monitor.c
apaan: As with previous discussion int he PR, it is possible that the unlock dialog has already been closed again before suspend, it the system was slow. This patch deliberately send show message after suspend as well, so that it will be shown again.

The callback signature mimics "PrepareForSleep" logind message, so it has prepare parameter (renamed from ```active``` for clarty). Maybe some time in the future we will need that for another use.

src/gs-listener-dbus.c
apaan: thanks

src/gs-listener-dbus.c
apaan: Thanks

apaan:
Force-pushed a new version of this patch:
- Fixed indentation problems (thanks @vkareh, @monsta)
- Fixed not setting ```logind_inhibit_lock``` to zero after closing it (thanks @monsta)
- Update commit message

src/gs-monitor.c
monsta: Ok, makes sense.

PR #205, opened 2019-08-16
Use gresources for mate-screensaver-preferences.ui

rbuj:


raveit65:
This works fine and mate-preferences dialog runs well, but why not adding `lock-dialog-default.ui` to gresource file too?

PR #207, opened 2019-10-04
Add keywords key to .desktop files

flexiondotorg:
Taken from downstream Debian patch.


rbuj:
Keywords key can be translated. Use `_Keywords=...` instead of `Keywords=`.

flexiondotorg:
Thanks @rbuj, changed as suggested :+1:

flexiondotorg:
Translator notes added.

raveit65:
Merged without deleting the branch!

PR #209, opened 2019-10-12
Travis CI: build mate-common from master branch

sc0w:


raveit65:
4 arguments not to merge this PR:
1. It's better to build mate-common from a specific version here and in all other repos.
Than we see it clear when we can remove those lines from travis config, when all distros use mate-common-1.24.0.
2. There is nothing missing in mate-common-1.23.3.tar.xz which we need for actual building.
3. There isn't an improvement in mate-common master which i needed for building.
4. Which build issue or report do you want to fix with this PR?

sc0w:
currently mate-menus is building in mate-screensaver from master branch?

why mate-common from tarball and mate-menus from master branch?

is very simply, the development is continuous, in my opinion mate-screensaver in master, needs to build with master

@mate-desktop/core-team please vote

rbuj:
I agree with @raveit65. From my point of view, autoconf-archive is not a direct dependency for mate-screensaver, building mate-common from tarball presumably will be superseded by adding autoconf-archive as a runtime dependency to mate-common. For that reason, I don't see that building mate-common from source as a permanent stage before building this package.

sc0w:
@rbuj  why mate-menus from git master and mate-common from tarball? why both aren't building in the same way?

rbuj:
@sc0w mate-common is a data package. You can also patch the distributed text files instead of building and installing the package from the source, but it's not necessary because you already have a release tarball. In any case, this is a temporary change and It will be superseded on next release.

sc0w:
There is no sense for me, to make one temporary change instead one permanently change

raveit65:
>  why mate-menus from git master and mate-common from tarball? why both aren't building in the same way?

Thank you for giving this example, because building mate-menus from git master is good example how bad it is not to use a fixed tarball for a temporary build fix.
We had a API breakage with mate-menus-1.21.0.
https://github.com/mate-desktop/mate-menus/commit/3a5d32495062ac2b63a679314b51e31c923bf25d
For this reasons we had to build all depending packages from git master or from 1.21.0 release.
But actual all distros use latest stable mate-menus-1.22.1 for travis builds.
See our distro with oldest package database.
https://packages.debian.org/search?keywords=mate-menus
SID or testing are using 1.22.1 too.
In other words we can remove building mate-menus from git from travis.yml

So, building from master for a temporary build fix is really counterproductive because it is easy to forget to remove those step, when it isn't needed any more.

rbuj:
PR https://github.com/mate-desktop/mate-screensaver/pull/211

sc0w:
(offtopic) I am wondering why ubuntu builds always very slowly.

raveit65:
> (offtopic) I am wondering why ubuntu builds always very slowly.

Yes , since we switch to use 19.10 ubuntu build is the slowest.
Maybe, because of downloading docker releases and packages from ubuntu 19.10 repos?

PR #210, opened 2019-10-12
[1.22] Travis CI: don't build mate-menus from git master

sc0w:


sc0w:
I added one more commit to fix this error in ubuntu build:

`Requested 'libmate-menu >= 1.21.0' but version of libmate-menu is 1.20.2`

There is no sense to have distro in 1.22 branch which hasn't 1.22 packages

PR #212, opened 2019-10-15
Migrate from intltool to gettext

yetist:


Makefile.am
```diff
@@ -25,22 +25,15 @@ EXTRA_DIST =			\
 	INSTALL			\
 	README			\
 	ChangeLog		\
+	its				\
```
raveit65: Position of back-slash can be improved, or not?

raveit65:
Comparing old and new *.pot file shows me that all translations are there.
@yetist 
Beside from the little formatting issue all looks good..

Makefile.am
yetist: Done.

raveit65:
I squashed formating fix in other commit.
Why an extra commit?

yetist:
> I squashed formating fix in other commit.
> Why an extra commit?

OK, Thanks.

PR #221, opened 2020-03-10
Travis CI: install mate-common 1.23.3 from tarball is no longer required

rbuj:


raveit65:
@lukefromdc 
I think it's better to wait for fedora 32 and ubuntu-20.04 release in a few weeks, to avoid editing travis configs.
See https://github.com/mate-desktop/mate-utils/pull/269 for more info.

PR #222, opened 2020-03-15
Remove function casting for g_idle_add (function, data)

rbuj:


raveit65:
You forgot to call reviewers.

PR #223, opened 2020-03-15
gs-lock-plug: remove function casting

rbuj:
`dialog_timed_out` and `response_cancel_idle_cb` are [GSourceFunc](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#GSourceFunc) functions which are not called outside `g_timeout_add`; for that reason, they don't need a function casting like `date_time_update`, which is called within `g_timeout_add` and directly `date_time_update (plug);`

### DIFF
--- a/src/gs-lock-plug.c
+++ b/src/gs-lock-plug.c
@@ -387,8 +387,10 @@ gs_lock_plug_response (GSLockPlug *plug,
 }
 
 static gboolean
-response_cancel_idle_cb (GSLockPlug *plug)
+response_cancel_idle_cb (gpointer user_data)
 {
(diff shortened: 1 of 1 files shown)


raveit65:
Why not calling reviewers?

monsta:
Interesting, with the correct function return type (`gboolean`) the casting via `(GSourceFunc)` doesn't cause a warning. That's casting of a pointer to some type to a pointer to `void` (in the function argument), same as done with `response_cancel_idle_cb` and `dialog_timed_out` in the same file.

rbuj:
@monsta done

PR #224, opened 2020-03-15
Custom background image for lock screen

### LINKED ISSUE #55: Option to change background of lock-screen
It would be really, really nice to see option to change background of lock-screen that MATE has. This feature would fit perfectly inside "Desktop Settings" window, accessible from Control Center :)
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3460377-option-to-change-background-of-lock-screen?utm_campaign=plugin&utm_content=tracker%2F1635706&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1635706&utm_medium=issues&utm_source=github).
</bountysource-plugin>

rbuj:
closes #55

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1112,7 +1112,6 @@ data/Makefile
 data/mate-screensavers.menu
 data/mate-screensaver.pc
 data/org.mate.ScreenSaver.service
-data/org.mate.screensaver.gschema.xml
 data/images/Makefile
 data/images/cosmos/Makefile
 savers/Makefile
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -46,6 +46,9 @@ dbussessionservice_DATA = org.mate.ScreenSaver.service
 gsettingsschema_in_files = org.mate.screensaver.gschema.xml.in
 gsettings_SCHEMAS = $(gsettingsschema_in_files:.xml.in=.xml)
 
+org.mate.screensaver.gschema.xml: org.mate.screensaver.gschema.xml.in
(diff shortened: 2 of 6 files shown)


milouse:
I agree with that PR. I have a simple workflow to explain it: I use a wallpaper manager, and each time my wallpaper changes, it creates a blurred copy of it, which is saved in an accessible place, to be used by LightDM. I would like my lock screen to use the same blurred picture, instead of my current background.

Thank you for your work!

milouse:
I confirm this patch runs fine after several days. No unexpected things.

Maybe an improvement: could you make the new background picture for lockscreen be desactivated when the lock screen is desactivated (previous checkbox)? The two options will look more linked together.

Capture du 2020-03-25 19-35-26

rbuj:
@milouse The screen can be locked in other ways, choosing System - Lock Screen from mate-panel classic menu.

Screenshot at 2020-03-25 20-23-17

milouse:
Ho, yes, you're right! You can forgot my last comment.

sc0w:
@milouse and you can do the same quikly with <kbd>ctrl</kbd><kbd>alt</kbd><kbd>L</kbd>

raveit65:
I just updated fedora 32 and rawhide with PR.
https://bodhi.fedoraproject.org/updates/FEDORA-2020-edc02ca65b
Thanks again.

PR #226, opened 2020-05-16
Fix invalid gnome-logo-white.svg

lah7:
This fixes an issue where the "Floating GNOME" screensaver in MATE 1.24 is blank as the SVG itself cannot be loaded. It wasn't possible to load the SVG in viewers or editors either.

The Floating GNOME screensaver is now operational after this change.


esteelpaz:
This issue affects my various MATE distros, so thanks to @lah7 for filing this . . . the GNOME Floating Feet screensaver is sorely missed.  Appreciate the fix on it . . . .

lukefromdc:
For some reason the floating foot logo still works with my somewhat old mate-screensaver locally built 1.24.0+git20200312-1luke1_amd64.deb package and presumably also with the package I built from the 1.24.0 release

lah7:
I would _guess_ whichever library parsing the SVG "just worked" for some versions/distros all these years. The [original SVG](https://github.com/mate-desktop/mate-screensaver/blob/0e33476937b0a5c9fd3892dfa9d9d3e4b215c02d/data/images/gnome-logo-white.svg?short_path=b8ee84d) comes up as invalid (the `xmlns` in particular) when thrown into the [W3C validator](https://validator.w3.org/check). GitHub, Inkscape and Eye of MATE were unable to render the file.

Over here, it's working again under Ubuntu MATE 20.04, where the bug was spotted.

PR #235, opened 2020-10-16
mate-screensaver-preferences: Show lock image preview on file chooser

rbuj:
Screenshot at 2020-10-16 12-44-14


raveit65 (approved):
Works like a charm.
Very nice new feature.

PR #238, opened 2020-12-08
lock-dialog-default: fix deprecations reported by glade

rbuj:


raveit65 (approved):
LGTM.
logout- dialog looks like before

PR #240, opened 2021-01-13
Removed translation for icon file name

Pjotr123:
Can't be corrected on Transifex....

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -38,7 +38,7 @@ msgstr "Voorkeuren voor de schermbeveiliging instellen"
 #: data/mate-screensaver-preferences.desktop.in:6
 #: src/mate-screensaver.desktop.in.in:7
 msgid "preferences-desktop-screensaver"
-msgstr "voorkeuren-bureaublad-schermbeveiliging"
+msgstr "preferences-desktop-screensaver"
 
 #. Translators: Search terms to find this application. Do NOT translate or
 #. localize the semicolons! The list MUST also end with a semicolon!


raveit65:
> Can't be corrected on Transifex....

Why?
And every pull from transifex will override your commit.

raveit65:
Hmm weird, `msgid "preferences-desktop-screensaver"` isn't at transifex.

raveit65:
Ok, this shouldn't be translated.
```
# Translators: Do NOT translate or transliterate this text (this is an icon file name)!
Icon=preferences-desktop-screensaver
```
https://github.com/mate-desktop/mate-screensaver/blob/master/src/mate-screensaver.desktop.in.in#L5

Seems we never pushed that to transifex.

raveit65:
We pushed it to transifex with that commit.
https://github.com/mate-desktop/mate-screensaver/commit/81564b7ce7a3c12811e87a8932689a7dd8268ae8#diff-18e2cbf5563493aa14ac82baf62c952e1b46ad3832dc7cd0781b4add197635c0R30
But it was translated by accidental with that pull from transifex https://github.com/mate-desktop/mate-screensaver/commit/1d404773691ff39da4fb2cb27084bf90385c22ab#diff-44a066469e7c75b3cdbbe18743cae1045124dcb86d20c65269ba4c323b04f61cR36

I don't get why i don't find that msgid at transifex.

Pjotr123:
Perhaps you can simply correct it in the Dutch .po file? Because it
looks like that correction won't be overwritten by the next sync with
Transifex, as the string isn't present on Transifex anymore....

2021-01-13 21:11 GMT+01:00, raveit65 <notifications@github.com>:

raveit65:
yeah, sure i will do.
I am bit busy these days, let me think about a while.
I 'd like to understand why the msgid is missing at transifex.

PR #244, opened 2021-01-17
mate-screensaver.directory.desktop: The Encoding key is deprecated

rbuj:
https://specifications.freedesktop.org/desktop-entry-spec/latest/apc.html


raveit65 (approved):
LGTM

PR #245, opened 2021-01-24
Corrected several icon translations (which broke the .desktop icons)

sebastianschauenburg:
This string should not be translated at all (and it's not available on Transifex for translation, I checked). Similar fix as https://github.com/mate-desktop/mate-screensaver/pull/240

### DIFF
--- a/po/ca.po
+++ b/po/ca.po
@@ -36,7 +36,7 @@ msgstr "Establiu les preferències de l'estalvi de pantalla"
 #: data/mate-screensaver-preferences.desktop.in:6
 #: src/mate-screensaver.desktop.in.in:7
 msgid "preferences-desktop-screensaver"
-msgstr "preferences-desktop-screensaver"
+msgstr ""
 
 #. Translators: Search terms to find this application. Do NOT translate or
 #. localize the semicolons! The list MUST also end with a semicolon!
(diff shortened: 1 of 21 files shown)


rbuj:
See #242

sebastianschauenburg:
Added fixes for several other languages as well.

sebastianschauenburg:
if there is any doubt about these changes, please read [my recent blogpost](https://www.schauenburg.nl/posts/2021/01/11_translators_-_please_do_not_translate_everything.html) about this specific enhancement

sebastianschauenburg:
@rbuj seems like #242 is OK and has been merged. Could you do the same with this request please? :-)

rbuj:
l10n: The translations are managed at https://www.transifex.com/mate/MATE/dashboard/.
We have already fixed this kind of issue on other MATE components, as we don't fetch the translation for the icon name.

sebastianschauenburg:
@rbuj I checked the strings at Transifex and they were not present there anymore, currently. But in this repository, the .po files still contain the translated (or copied) strings, which seem incorrect? I'm fine with not accepting this pull request, but it would mean #242 should've also not been accepted, for consistency :wink:  Just tried to help out here.

rbuj:
Even if the icon names are in the catalog they are not used for generating the desktop files.

PR #250, opened 2021-05-08
Change date format for en_GB

JohnVeness:
UK date formats usually have the day of the month before the name of the month


raveit65:
We are using transifex for translations, please become a member of your language group an fix it there.
https://www.transifex.com/mate/MATE/

Issue #100, opened 2016-09-02
Feature: Be able to adjust clock (time) of the locked screen


ThanosApostolou:
Even thought I've set my Clock Preferences at 24 hour format, the locked screen shows 12 hour format. The clock panel item and login screen (lightdm greeter) work normally. I asked the question here: https://ubuntu-mate.community/t/locked-screen-shows-12-hour-format-even-if-my-clock-properties-are-set-for-24-hour-format/8619/1 and they told me that the hour format is intergrated into the code. Can you make it adjustable by the clock preferences?
##

richwales:
I managed to fix this by adding the following line to /etc/default/locale:

PR #252, opened 2021-06-06
Allow change of time and date formats on lock screen.

eugenesan:
This change allows modification of default time and date format defined by current locale.
Closes https://github.com/mate-desktop/mate-screensaver/issues/100

Default ('local') settings maintains old behavior.
User is able to use any format defined by https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format.

Examples:
%R for time means 24H format.
%F for date means YYYY-MM-DD (ISO 8601 format).

### DIFF
--- a/src/gs-lock-plug.c
+++ b/src/gs-lock-plug.c
@@ -58,6 +58,10 @@
 
 #define KEY_LOCK_DIALOG_THEME "lock-dialog-theme"
 
+#define KEY_LOCK_DIALOG_T_FMT "lock-dialog-time-format"
+#define KEY_LOCK_DIALOG_D_FMT "lock-dialog-date-format"
+
+
 #define MDM_FLEXISERVER_COMMAND "mdmflexiserver"
 #define MDM_FLEXISERVER_ARGS    "--startnew Standard"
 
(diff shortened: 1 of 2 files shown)


raveit65 (approved):
LGTM,
nice change and it does what it promise.

mbkma (changes requested):
- tfmt, dfmt need to be of type gchar (see https://developer.gnome.org/gio/stable/GSettings.html#g-settings-get-string) 
- You need to free tfmt and dfmt with g_free
- You need to free gchar* when overwriting them
- You should use g_strcmp0 (https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strcmp0) instead of strcmp (it is null safe)

eugenesan:
>     * tfmt, dfmt need to be of type gchar (see https://developer.gnome.org/gio/stable)/GSettings.html#g-settings-get-string)
> 
>     * You need to free tfmt and dfmt with g_free
> 
>     * You need to free gchar* when overwriting them
> 
>     * You should use g_strcmp0 (https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strcmp0) instead of strcmp (it is null safe)

Oppsie, silly me.
I was in such a rush that I forgot we are in "gnome" environment and stuff needs to be released ;-)
I am sure going to apply all your recommendations.

Also, I wonder if I can skip checking return value of g_date_time_format()?
Looks like it always outputs proper string.
If the parsing fails, it just outputs input string.
Worst case scenario the user will see that and correct.

raveit65:
@mbkma 
Are your concerns solved?

src/gs-lock-plug.c
```diff
@@ -293,12 +297,42 @@ date_time_update (GSLockPlug *plug)
...
 	datetime = g_date_time_new_now_local ();
-	time = g_date_time_format (datetime, "%X");
-	/* Translators: Date format, see https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format */
-	date = g_date_time_format (datetime, _("%A, %B %e"));
+	if (g_strcmp0(tfmt, "locale") == 0)
+	{
+		// Use locale default format
+		time = g_date_time_format (datetime, "%X");
+	}
+        else
```
mbkma, line 320: indent (`else` should be aligned with `if`)

src/gs-lock-plug.c
```diff
@@ -293,12 +297,42 @@ date_time_update (GSLockPlug *plug)
...
+		// Apply user defined format
+		time = g_date_time_format (datetime, tfmt);
+	}
+
+	if (g_strcmp0(dfmt, "locale") == 0)
+	{
+		// Use locale default format
+		date = g_date_time_format (datetime, _("%A, %B %e"));
+	}
+        else
```
mbkma, line 331: indent (`else` should be aligned with `if`)

src/gs-lock-plug.c
```diff
@@ -293,12 +297,42 @@ date_time_update (GSLockPlug *plug)
 	GDateTime *datetime;
 	gchar *time;
 	gchar *date;
+	gchar  *tfmt;
```
mbkma, line 300: indent (`*tfmt` and `*dfmt` should be aligned with `*date`)

mbkma (commented):
besides a few indent issues it is good :)

raveit65:
@eugenesan 
Any chance to fix this indent issues?

eugenesan:
@raveit65 
Sorry for the delay (I was away).
I've pushed requested changes.

raveit65 (approved):
Intends issues are fixed and screensaver runs fine.

mbkma (approved):
thanks for contributing, it works as expected

PR #281, opened 2023-05-13
configure.ac: The macro 'AC_HELP_STRING' is obsolete

raveit65:
same as https://github.com/mate-desktop/mate-screensaver/pull/278


lukefromdc (approved):
This configures and builds without issue, and ./configure --help works fine. This has been supported well over a decade so no distro that will ever ship MATE 1.28 will have an issue from this, and we've done this in most of the other packages by now. Will merge

PR #296, opened 2024-05-16
Cleanups following #293

cwendling:
* removes spurious leading dashes in UI file
* use a grid with variable column widths instead of 4 columns with manual ¼/¾
* align labels to the left
* add mnemonic and tooltip to the lock-delay label


zhuyaliang (approved):
The compilation runs well, and the correction UI

PR #298, opened 2024-05-21
mate-screensaver-preferences: Fix lock-delay time wording

cwendling:
The scale controls the delay after which to lock the screen once the screensaver activated, not whether or not it happens -- which has its own checkbox.  So remove the "Never" wording for value 0, which was inherited from an earlier version of the activate-delay scale that is not used anymore (the activate-delay scale is now 1-120, never 0).


vkareh (approved):
LGTM

PR #299, opened 2024-05-21
mate-screensaver-preferences: Simplify mnemonic widget association

cwendling:
Instead of doing the binding in the code, do it in the UI file.

This should not change anything but cleaning up the code a bit.


lukefromdc (approved):
This is a simple change that also makes the code more internally consistant. No build issues, and running in my default wayland session I was able to open mate-screensaver-preferences by forcing x11 use (via xwayland) and it seemed to work though of course keyboard control is a bit cranky that way. In the middle of a complex session so logging out for a full test could not be done right now.

vkareh (approved):

PR #300, opened 2024-06-21
Listen to org.freedesktop.ScreenSaver events

### LINKED ISSUE #57: MATE ScreenScreensaver should implement "XDG" Idle Inhibition Service specification
GNOME and KDE already implement it and at least VLC and Firefox (33+) make use of it. It's not an "offical" Freedesktop specification but it has become a "defacto-standard".
Link: http://people.freedesktop.org/~hadess/idle-inhibition-spec/

Things to do:
1. Claim DBus bus name `org.freedesktop.ScreenSaver` additionally to `org.mate.ScreenSaver`
2. Expose the MATE ScreenSaver interface (`org.mate.ScreenSaver`) as `org.freedesktop.ScreenSaver` on the object paths `/ScreenSaver` and `/org/freedesktop/ScreenSaver`
3. That's it! MATE ScreenSaver already implements the required methods: `Inhibit` and `UnInhibit` correctly!

The reason why you have add both the `/ScreenSaver` and `/org/freedesktop/ScreenSaver` object paths is because KDE orginally only added `/ScreenSaver` while `/org/freedesktop/ScreenSaver` fits into the official scheme. The spec fails to mention object paths entirly, but [GNOME adds both paths](https://git.gnome.org/browse/gnome-settings-daemon/tree/plugins/screensaver-proxy/gsd-screensaver-proxy-manager.c#n45) so I guess that is what should be used...
Also GNOME exports all for its ScreenSaver methods via `org.freedesktop.ScreenSaver` althrough the spec only says you have to export the `Inhibit` and `UnInhibit` methods so I guess that's how it should be done...
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4363731-mate-screenscreensaver-should-implement-xdg-idle-inhibition-service-specification?utm_campaign=plugin&utm_content=tracker%2F1635706&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1635706&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oz123:
This fixes #57.

### DIFF
--- a/src/gs-listener-dbus.c
+++ b/src/gs-listener-dbus.c
@@ -51,9 +51,9 @@ static DBusHandlerResult gs_listener_message_handler    (DBusConnection  *connec
         DBusMessage     *message,
         void            *user_data);
 
-#define GS_LISTENER_SERVICE   "org.mate.ScreenSaver"
-#define GS_LISTENER_PATH      "/org/mate/ScreenSaver"
-#define GS_LISTENER_INTERFACE "org.mate.ScreenSaver"
+#define GS_LISTENER_SERVICE   "org.freedesktop.ScreenSaver"
+#define GS_LISTENER_PATH      "/org/freedesktop/ScreenSaver"
+#define GS_LISTENER_INTERFACE "org.freedesktop.ScreenSaver"
 
 /* systemd logind */
 #define SYSTEMD_LOGIND_SERVICE   "org.freedesktop.login1"
@@ -1370,7 +1370,7 @@ do_introspect (DBusConnection *connection,
 
 	/* ScreenSaver interface */
 	xml = g_string_append (xml,
-	                       "  <interface name=\"org.mate.ScreenSaver\">\n"
+	                       "  <interface name=\"org.freedesktop.ScreenSaver\">\n"
 	                       "    <method name=\"Lock\">\n"
 	                       "    </method>\n"
 	                       "    <method name=\"Unlock\">\n"
--- a/src/mate-screensaver-command.c
+++ b/src/mate-screensaver-command.c
@@ -34,9 +34,9 @@
 #include <dbus/dbus-glib.h>
 #include <dbus/dbus-glib-lowlevel.h>
 
-#define GS_SERVICE   "org.mate.ScreenSaver"
-#define GS_PATH      "/org/mate/ScreenSaver"
-#define GS_INTERFACE "org.mate.ScreenSaver"
+#define GS_SERVICE   "org.freedesktop.ScreenSaver"
+#define GS_PATH      "/org/freedesktop/ScreenSaver"
+#define GS_INTERFACE "org.freedesktop.ScreenSaver"
 
 static gboolean do_quit       = FALSE;
 static gboolean do_lock       = FALSE;


spixi:
This would also require fixes in mate-power-manager and mate-session-manager.
https://github.com/mate-desktop/mate-power-manager/blob/937ef297a9b1552c044958e69e10183aa658e5bf/src/gpm-screensaver.c#L35
https://github.com/mate-desktop/mate-session-manager/blob/fa1bf6c2fb45ff0263465537a0730ebff2b2010d/mate-session/gsm-manager.c#L95

lukefromdc:
I don't really have a good way to test this given both my setup and workflow

oz123:
I can build deb packages for you if it helps.
Testing is very easy. Start netflix or youtube in your browser, setup screensaver to start after 2 minutes. 
You will notice that despite streaming video, the screensaver starts after 2 minutes.
When installing the packages with the patch,  the screensaver will not start if you play video.

lukefromdc:
I do not have ANY streaming accounts but when playing video in things like MPV I've never had trouble
with the screensaver. HOWEVER, I do not have the available bandwidh (no landline) for long format video
in browser (from a remote source) plus I normally set screensavers for very long timeouts, so this could
exist but never trigger in my workflow.

Building these is not the problem for me to test, not at all. Rather I have to shut down all my other work,
switch from the mate-wayland session back to the default xorg session, reconfigure my screensaver 
settings, and then do something like playing a local video in the browser.

oz123:
Thanks for the detailed explanation. As it seems, you are currently the only member with merge rights, so it's up to you. 
I am able to set a virtual machine for you to connect with a VNC so you can test this.

lukefromdc:
I will NOT be able to do a remote VM connection, never done anything like that.
If others test this and get good results I can still merge it, thankfully.

This is another example of why I cannot solo maintain MATE, we need to get more
people on the team but I don't know how to do that, determine who can be trusted
etc.

mmueller:
I've had this bookmarked for a while, and I just noticed the request for testers. I tried this on my system (Manjaro 24.2.1), built mate-common and mate-screensaver in ~/mate and ran mate-screensaver manually out of there. All of my other MATE packages are running out of the OS's maintained versions.

It works, but there is a quirk.

* I set screensaver to kick on after 1 minute.
* Power management is set to turn off display after 30 minutes.
* If I idle for a minute, screensaver kicks on (fade to black then lock).
* I confirmed that with my system's mate-screensaver, it still fades to black during the video and locks.
* I killed the system mate-screensaver and ran my local build.
* Now, if I idle while watching a Youtube video, screensaver _does not_ kick in. (Nice!)
* However, the quirk: If I pause the video and start doing anything else, about 3 seconds later, the screensaver activates. No fade out, just sudden black/locked screen.

I tested pausing with the trackpad and clicking, and using the keyboard (spacebar) and then typing away in another window, and it will literally lock the screen on me while I'm scrolling or typing.

Since I'm clicking and typing, I would expect the session to be considered active, not idle. I wonder if there's some race condition between "session is idle" flag being updated and the "screensaver is being suppressed right now" message from dbus?

It's not really the fault of this PR, but it's probably exposing something that has been wrong for a while. I'm curious, @oz123 do you see this behavior, too? Or maybe it's specific to something about my setup.

lukefromdc:
THANK YOU for testing! We now know this needs work before I can merge it. The bug you found is a nasty one and if it got into a release would require an emergency fix and re-release

spixi:
Please note, that we still have to listen for org.mate.ScreenSaver, because many third-party application will check for the DESKTOP_SESSION variable and will only send the org.mate.ScreenSaver events if they detect MATE.

oz123:
> Please note, that we still have to listen for org.mate.ScreenSaver, because many third-party application will check for the DESKTOP_SESSION variable and will only send the org.mate.ScreenSaver events if they detect MATE.

Can you be more specific? Which many ?

Firefox\Chrome is not anymore.

spixi:
> > Please note, that we still have to listen for org.mate.ScreenSaver, because many third-party application will check for the DESKTOP_SESSION variable and will only send the org.mate.ScreenSaver events if they detect MATE.
> 
> Can you be more specific? Which many ?
> 
> Firefox\Chrome is not anymore.

You can blame me that I introduced this into PCSX2. Other projects can be found here (and this is only the list of projects which host their code on GitHub):
https://github.com/search?q=org.mate.ScreenSaver+DESKTOP_SESSION&type=code

oz123:
Oh. Crap. All those should just list to `org.freedesktop.ScreenSaver`. I can give it a try to listen on multiple event types.

@mmueller Unfortunately, I can't confirm the behavior you are describing...

spixi:
> Oh. Crap. All those should just list to `org.freedesktop.ScreenSaver`. I can give it a try to listen on multiple event types.
> 
> @mmueller Unfortunately, I can't confirm the behavior you are describing...

Sometimes, the logic is more hidden, but yeah, we now have a chicken egg problem and should listen to both events. https://github.com/videolan/vlc/blob/5857e58efdf5aad7c20bbe19be897f2865053174/modules/misc/inhibit/dbus.c

vkareh (changes requested):
This is a good change, but it will break a lot of software that expects `org.mate.ScreenSaver` when running on MATE.

We should keep this code intact and introduce a new dbus call/listener for `org.freedesktop.ScreenSaver`.

oz123:
Can you give a specific example? I'm using this and maybe I'm missing something.

oz123:
Yes. VLC works fine as it relies also org.free.desktop.
We can't look for all possible repos. IMO any software that relies on org.mate.ScreenSaver, b should just listen to org.freedesktop.ScreenSaver, it's de facto the standard.

vkareh (approved):
/lgtm

Can you also send patches for the various mate packages that use `org.mate.ScreenSaver`?
- https://github.com/mate-desktop/mate-notification-daemon/blob/master/src/daemon/daemon.c#L1152-L1154
- https://github.com/mate-desktop/mate-power-manager/blob/master/src/gpm-screensaver.c#L35-L37
- https://github.com/mate-desktop/mate-session-manager/blob/master/mate-session/gsm-presence.c#L39-L41
- https://github.com/mate-desktop/mate-settings-daemon/blob/master/plugins/smartcard/msd-smartcard-plugin.c#L48-L50

vkareh:
@cwendling 
> So, my question is whether we should keep `org.mate.ScreenSaver` with all methods, and implement `org.freedesktop.ScreenSaver` _as well_

That would be the correct way, in my opinion, of handling this type of change. We need (like the #57 says) to claim _both_ bus names. Which, considering that applications do things like `if XDG_DESTKOP == MATE { busname="org.mate.ScreenSaver" }`, if we don't do that, they will break.

PR #307, opened 2025-04-24
Fix mate-screensaver may not respond on logind dbus commands

### LINKED ISSUE #306: mate-screensaver ignore loginctl msgs if it's started by days
# Environment

DM: lightdm 1.30.0, gdm3: 47.0 (Tested on both)
OS: Ubuntu 24.04
MATE: 1.26.2

# Description

Hi! I discovered some bug: if mate-screensaver is stated by dbus (using `/usr/share/dbus-1/services/org.mate.ScreenSaver.service`), then it ignore commands from `loginctl` command (`lock-session`/`unlock-session`).

I noticed this because another interesting behaviour: `mate-screensaver` could be started using autostart and dbus. Both of them try to start mate-screensaver at the same time and there is no guarantee who start it first. On some system (Alt Linux 10.4) I got situation when mate-screensaver is started by dbus in most cases. On other distros -- it's started in most cases by autostart.

But, as I said. There is no guarantee who start it first. You can easily check it this way:

1. replace` Exec` command in `/usr/share/dbus-1/services/org.mate.ScreenSaver.service` and `/etc/xdg/autostart/mate-screensaver.desktop` on something like this:
```ini
Exec=/sbin/sh -c 'sleep n && /usr/bin/mate-screensaver'
```
2. try to set different sleep time in dbus and autostart files. So, you could ensure the the one, who has lowest sleep time will be started started
3. You can check it that way: run `pstree` and check the parent of mate-screensaver and it args

So, I believe that there is no problem in this race, if dbus would start mate-screensaver correctly, but it doesn't

# How to reproduce

So, you could reproduce the problem this way

```bash
sudo sed -i 's/Exec=.*/Exec=sh -c "sleep 5 \&\& mate-screensaver"/g' /etc/xdg/autostart/mate-screensaver.desktop
sudo reboot
# login by any user
loginctl lock-session # not working
```

lo1ol:
GetSessionByPID may fail in some cases. For example when it was started by D-bus.

Let's try to additionally get session D-bas path using GetSession method, like loginctl doint it

For more info read: https://github.com/mate-desktop/mate-screensaver/issues/306

### DIFF
--- a/src/gs-listener-dbus.c
+++ b/src/gs-listener-dbus.c
@@ -2483,6 +2483,7 @@ query_session_id (GSListener *listener)
 	DBusError       error;
 	DBusMessageIter reply_iter;
 	char           *ssid;
+	const char* auto_session_id;
 
 	if (listener->priv->system_connection == NULL)
 	{
@@ -2498,8 +2499,9 @@ query_session_id (GSListener *listener)
 	if (listener->priv->have_systemd) {
 		dbus_uint32_t pid = getpid();
 
+		/* first try to get session D-bus path by PID */
 		message = dbus_message_new_method_call (SYSTEMD_LOGIND_SERVICE,
-		                                        SYSTEMD_LOGIND_PATH,
+							SYSTEMD_LOGIND_PATH,
 							SYSTEMD_LOGIND_INTERFACE,
 							"GetSessionByPID");
 		if (message == NULL)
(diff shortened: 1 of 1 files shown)


lo1ol:
you can reproduce problem this way:

```bash
pkill mate-screensaver
mate-screensaver --debug &
loginctl lock-session # this command will be ignored on old mate-screensaver
```

src/gs-listener-dbus.c
```diff
@@ -2522,19 +2523,60 @@ query_session_id (GSListener *listener)
...
+		}
+
+		gs_debug ("%s raised:\n %s\n\n", error.name, error.message);
+		dbus_error_free (&error);
+
+        // if getting session D-bus path by PID failed, try to get it using session Id
+		message = dbus_message_new_method_call (SYSTEMD_LOGIND_SERVICE,
+		                                        SYSTEMD_LOGIND_PATH,
+							SYSTEMD_LOGIND_INTERFACE,
+							"GetSession");
```
vkareh, line 2544: Indentation is inconsistent here

src/gs-listener-dbus.c
```diff
@@ -2522,19 +2523,60 @@ query_session_id (GSListener *listener)
...
+			gs_debug ("Couldn't allocate the dbus message");
 			return NULL;
 		}
 
-		dbus_message_iter_init (reply, &reply_iter);
-		dbus_message_iter_get_basic (&reply_iter, &ssid);
+        // pass 'auto' as argument
+        // loginctl do it this way
+        // https://github.com/systemd/systemd/blob/10594762c7c26f944d038ed87258d997eb98079c/src/login/loginctl.c#L1066
+		const char* autoSessionId = "auto";
```
vkareh: Use snake-case, like in the rest of the code: `auto_session_id`. Also move the declaration to the top of the function block (but still assign it here).

src/gs-listener-dbus.c
```diff
@@ -2522,19 +2523,60 @@ query_session_id (GSListener *listener)
...
+		if (dbus_message_append_args (message,
+					      DBUS_TYPE_STRING,
+					      &autoSessionId, DBUS_TYPE_INVALID) == FALSE)
+		{
+			gs_debug ("Couldn't add args to the dbus message");
+			return NULL;
```
vkareh, line 2562: You have a memory leak here. Need to unref `message`.

src/gs-listener-dbus.c
```diff
@@ -2522,19 +2523,60 @@ query_session_id (GSListener *listener)
...
+		}
 
-		dbus_message_unref (reply);
+		/* FIXME: use async? */
+		reply = dbus_connection_send_with_reply_and_block (listener->priv->system_connection,
```
vkareh, line 2566: You might need a null check on this

vkareh (changes requested):

src/gs-listener-dbus.c
lo1ol: f

src/gs-listener-dbus.c
lo1ol: f

src/gs-listener-dbus.c
lo1ol: f

src/gs-listener-dbus.c
lo1ol: Not sure about this. [dbus manual says](https://dbus.freedesktop.org/doc/api/html/group__DBusConnection.html#ga8d6431f17a9e53c9446d87c2ba8409f0)

> If a normal reply is received, it is returned, and removed from the incoming message queue. If it is not received, [NULL](https://dbus.freedesktop.org/doc/api/html/group__DBusMacros.html#ga070d2ce7b6bb7e5c05602aa8c308d0c4) is returned and the error is set to [DBUS_ERROR_NO_REPLY](https://dbus.freedesktop.org/doc/api/html/group__DBusProtocol.html#gade2cb1c5a6adf47af18672865f233b6d). If an error reply is received, it is converted to a [DBusError](https://dbus.freedesktop.org/doc/api/html/structDBusError.html) and returned as an error, then the reply message is deleted and [NULL](https://dbus.freedesktop.org/doc/api/html/group__DBusMacros.html#ga070d2ce7b6bb7e5c05602aa8c308d0c4) is returned. If something else goes wrong, result is set to whatever is appropriate, such as [DBUS_ERROR_NO_MEMORY](https://dbus.freedesktop.org/doc/api/html/group__DBusProtocol.html#gac32eaf0b92f798307853cd4fe0cf11c2) or [DBUS_ERROR_DISCONNECTED](https://dbus.freedesktop.org/doc/api/html/group__DBusProtocol.html#gaef6443e0bd8cdcc1606d985a775d07a7).

> Returns
> the message that is the reply or [NULL](https://dbus.freedesktop.org/doc/api/html/group__DBusMacros.html#ga070d2ce7b6bb7e5c05602aa8c308d0c4) with an error code if the function fails.

So, error should be always set on returned NULL. Also, there are 5 other places with `dbus_connection_send_with_reply_and_block` call and there are on check of returned value too

vkareh: You're right, thanks for pointing that out. Works for me.

PR #308, opened 2025-04-25
Fix warning due to source being removed twice

andrew-sayers:
Every time mate-screensaver is displayed, I get a message like `Source ID N was not found when attempting to remove it`.  Source IDs are reused over time, so this could potentially lead to a more serious issue in some cases.

This silences that warning, and removes the risk of the more serious issue.


lukefromdc:
I pushed a fix for master for some of the older travis build errors. On that commit the build finished but the deploy section failed again with
```
Install deployment dependencies
Authenticate deployment
no implicit conversion of nil into String (TypeError)
failed to deploy
```
which I do not know how to fix

vkareh (approved):

PR #317, opened 2026-03-11
fix memleak

mbkma:


vkareh (changes requested):
You need to unref `builder` in the error return path as well (above, where `gtk_builder_add_from_file` fails).

PR #325, opened 2026-06-30
Proposed solution for issue #324

### LINKED ISSUE #324: Behavior of "enable_checkbox" and "lock_checkbox" widgets when using dconf locks
### Steps to Reproduce

Here are the steps to reproduce the two issues.
1. Create a file in /etc/dconf/db/local.d/locks/ with the key /org/mate/screensaver/lock-enabled and run dconf update
2. Launch mate-screensaver-preferences

### Expected Behavior

First issue: 
Locking /org/mate/screensaver/lock-enabled should not lead to the locking of the enable_checkbox widget. 
Locking /org/mate/screensaver/idle-activation-enabled should lead to the locking of the enable_checkbox widget.

Second issue: 
Locking /org/mate/screensaver/lock-enabled should lead to the locking of the lock_checkbox widget.

### Actual Behavior

First issue:
The enable_checkbox widget is insensitive, while /org/mate/screensaver/lock-enabled is locked in dconf (this widget is linked with the /org/mate/screensaver/idle-activation-enabled key). 
At the same time, locking /org/mate/screensaver/idle-activation-enabled does not lead to locking enable_checkbox, although it should. 
This is how the mate-screensaver-preferences window looks when /org/mate/screensaver/lock-enabled is specified in a file within the /etc/dconf/db/local.d/locks/ directory.
Image

Second issue:
The lock_checkbox widget is still sensitive, while /org/mate/screensaver/lock-enabled is locked in dconf.

### Error Output

```shell

```

### Version

1.28

### Distribution

etc

### Additional Context

Regarding the first issue: it appears this error arose during the copy-pasting of the config_get_enabled function code. 
The sensitivity of the enable_checkbox widget is set based on the is_writable value in the init_capplet function. However, inside config_get_enabled, is_writable is assigned a value checked against a different key (KEY_LOCK - "lock-enabled") instead of the correct key (KEY_IDLE_ACTIVATION_ENABLED - "idle-activation-enabled"). 
I applied https://github.com/mate-desktop/mate-screensaver/commit/af1071c1abb749657bdb97cc590aa0ccbc58b3f8 and it solved the problem described as first issue. 

But locking /org/mate/screensaver/lock-enabled still does not lead to the lock_checkbox widget being locked.

As I understand it, this is due to how the code determines whether the lock_checkbox widget is sensitive or not. The system setting "disable-lock-screen" (KEY_LOCK_DISABLE) is checked for this widget using the config_get_lock_disabled function; however, the ui_disable_lock function sets the widget to a sensitive state without verifying whether the "lock-enabled" key is writable. 
At the moment, I haven't come up with anything better than applying https://github.com/Mejioy/mate-screensaver/commit/75f96219fbea00552f35c972e9e4198904a49c64. I don't consider this the best solution; for instance, in the ui_set_enabled function, config_get_lock is called before ui_disable_lock to check the "lock-enabled" key and set sensitivity for the widget, but this gets overwritten by the ui_disable_lock implementation without this commit.

### Reproducibility

Every time

### Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have provided all the requested information above

Mejioy:


src/mate-screensaver-preferences.c
```diff
@@ -1117,9 +1117,14 @@ static void
...
 	GtkWidget *widget;
+	gboolean   is_writable;
 
 	widget = GTK_WIDGET (gtk_builder_get_object (builder, "lock_checkbox"));
-	gtk_widget_set_sensitive (widget, !disable);
+	config_get_lock (&is_writable);
+	if (is_writable)
+	{
+		gtk_widget_set_sensitive (widget, !disable);
+	}
```
vkareh: Wrapping the entire `gtk_widget_set_sensitive` call means that it gets skipped even if `disable == TRUE`.

I'd use `gtk_widget_set_sensitive (widget, is_writable && !disable);` to avoid the regression.

vkareh (changes requested):

Mejioy (commented):
Correction according to the comment from vkareh to avoid regression

